PR #109, opened 2014-04-09
Adding composer.json to publish on packagist

zardo:
The primary requirement to publish on packagist is add an composer.json file on root.


pelex:
Can we use RobLoach/component-installer?
https://github.com/RobLoach/component-installer#creating-a-component
It's copying all necessary files to web public folder after composer install or update.

PR #114, opened 2014-04-23
Add .slider('getAttribute', [attribute]) 

raulfortes:
Add .slider('getAttribute', [attribute])  Get the slider's [attributes](#options)

### DIFF
--- a/spec/ElementDataAttributesSpec.js
+++ b/spec/ElementDataAttributesSpec.js
@@ -91,13 +91,6 @@ describe("Element Data Attributes Tests", function() {
     expect(isRangeSlider).toBeFalsy();
   });
 
-  it("get attribute", function() {
-    slider = $("#maxSlider").slider();
-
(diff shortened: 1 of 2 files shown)


capegreg:
Looking for example to get single attribute instead of attribute object.

// this works but  I only need the enabled value attribute
var deploy_atts = $("#deployedMilDays").slider('getAttribute')
console.log(deploy_atts.enabled)

// getAttribute('enabled') ??

seiyria:
@capegreg according to the documentation that should work: https://github.com/seiyria/bootstrap-slider#functions

capegreg:
Thanks. 

Checking .slider('isEnabled') works for my needs. I was not able to check the slider enable state using getAttribute because enable was always true even when the slider state was changed using .slider("disable") or .slider("enable").

// working
$("#deployedDays").slider("disable")
$("#deployedDays").slider('isEnabled')

PR #243, opened 2014-10-24
Fix range-slider bug (index.html Example 2) where hidden but touchable tooltip-max moves up too high.

inamiy:
tooltip-max-bug

Above image is bug-reproducing-demo from e4ec897 which I just set `tooltip.opacity=1` forcefully.

In current version, these tooltips are draggable **even when they are invisible**.
I think it is OK for `tooltip-main (center)` and `tooltip-min` to be draggable on hidden because they are very near slider UI, but not torelable for `tootip-MAX`, so I fixed `tootip-MAX`'s top position.

### DIFF
--- a/index.html
+++ b/index.html
@@ -138,6 +138,10 @@
 		#R, #G, #B {
 			width: 300px;
 		}
+
+        .tooltip {
+            opacity: 1 !important;
+        }
     </style>
 	<script type='text/javascript' src="bower_components/modernizr/modernizr.js"></script>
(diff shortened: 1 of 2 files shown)


PopeOnARope:
Hey wondering if this has been fixed? I am noticing the same problem in the most up-to-date version, but have managed to fix temporarily just by setting line 788 to this:
` this.tooltip_max.style.top = 0 + 'px'; `
screen shot 2015-01-21 at 10 07 23 am

seiyria:
It hasn't yet been merged, no.

PR #276, opened 2014-12-30
fix: js-accessible value was not changed by slider function

netzkollektiv:
The .value property of the element was not changed by the _setDataVal function. Therefore it was not possible to get the correct input's value by accessing the element.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1098,6 +1098,7 @@
 				var value = "value: '" + val + "'";
 				this.element.setAttribute('data', value);
 				this.element.setAttribute('value', val);
+                this.element.value = val;
 			},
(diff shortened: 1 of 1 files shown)


netzkollektiv:
Here you go: http://jsfiddle.net/t9ao6yht/

And here some explanations: http://stackoverflow.com/questions/6799780/in-javascript-difference-input-value-val-versus-input-setattributevalue-val

seiyria:
Can you please create a test or two with this behavior, simply showing this error in action?

rovolution:
@netzkollektiv if you add an automated test for this to our ```  /test/specs/PublicMethodsSpec.js ``` directory that would be perfect!

rovolution:
@seiyria I created a small test case for this. Will merge and then add my test

PR #282, opened 2015-01-13
outsource rules and variables

netzkind:
so it is possible to compile the slider from outside of this project without having ones variables overwritten by bootstrap-defaults and/or without errors for having bootstrap someplace other than specified here


rovolution:
@netzkind thanks for the suggestion! Since there have been a lot of recent changes to the LESS file, I went ahead and just made this change to master

Issue #286, opened 2015-01-19
_offset computation does not take into account css transforms of parents


sebtoun:
When a parent is positionned using css transform property, it's offset is not taken into account when computing offset property (_offset method). As a result, sliders handler are not correctly positionned. Particularly it prevents the use of both iScroll library and bootstrap-slider as iScroll use css transform if available to implement smooth scroll.

sebtoun:
[Here](http://jsfiddle.net/zefkc1rw/) is a jsfiddle showing the issue: scroll a bit and try to move the slider.

seiyria:
Is this fixed now, with your PR #286?

PR #287, opened 2015-01-19
Fix for issue #286: Accounts for translate css transform in _offset function

sebtoun:
Fixes issue #286. Adds translate css transform component in parent offset computation in _offset method. This particularily solves issues when using both iScroll and bootstrap-sliders as iscroll use css transforms to move scrollables around.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1170,10 +1170,17 @@
 			_offset: function (obj) {
 				var ol = 0;
 				var ot = 0;
+				var tr, mat;
 				if (obj.offsetParent) {
 					do {
-					  ol += obj.offsetLeft;
-					  ot += obj.offsetTop;
+						ol += obj.offsetLeft;
+						ot += obj.offsetTop;
+						tr = $(obj).css("transform");
+						if (tr !== "none"){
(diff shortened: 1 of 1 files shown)


seiyria:
Hm. Any chance you could add a test for this behavior?

sebtoun:
[Here](http://jsfiddle.net/zefkc1rw/) is a jsfiddle showing the issue: scroll a bit and try to move the slider.

sebtoun:
And [here](http://jsfiddle.net/zefkc1rw/) the patch in effects. Notice the slider is correctly moved even after the page has been scrolled.

sebtoun:
I'm afraid I won't be able to add any tests.

seiyria:
Sure, no worries. @rovolution - what do you think about the fix?

rovolution:
I think its good. As long as our existing test suite is passing, i dont see a need to add a new test for this (since its not really adding new functionality or exposing a new public method).

merging in

mpolyak:
Hi,

This change introduces an exception: Uncaught TypeError: object is not a function

On "tr = $(obj).css("transform");" when using bootstrap-slider without jQuery.

PR #296, opened 2015-01-29
(feat) Add tick marks

MelleB:
Add tick marks to the slider. Two new configuration options have been
defined:
- ticks (Array): used to define the values of the ticks. Overwrites
  min and max options.
- snap (int): Used to define snapping behaviour near tick marks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | formatter |	function |	returns the plain value |	formatter callback. Return the value wanted to be displayed in the tooltip |
 | natural_arrow_keys | bool | false | The natural order is used for the arrow keys. Arrow up select the upper slider value for vertical sliders, arrow right the righter slider value for a horizontal slider - no matter if the slider was reversed or not. By default the arrow keys are oriented by arrow up/right to the higher slider value, arrow down/left to the lower slider value. |
 | ticks | array | [] | Used to define the values of ticks. Tick marks are indicators to denote special values in the range. This option overwrites min and max options. |
+| tick_labels | array | [] | Defines the labels below the tick marks. Accepts HTML input. |
 | snap | float | 0 | Used to define the bounds of a tick. Snaps to the tick if value is within these bounds. |
 
 Functions
--- a/README.md
+++ b/README.md
@@ -126,8 +126,8 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | formatter |	function |	returns the plain value |	formatter callback. Return the value wanted to be displayed in the tooltip |
 | natural_arrow_keys | bool | false | The natural order is used for the arrow keys. Arrow up select the upper slider value for vertical sliders, arrow right the righter slider value for a horizontal slider - no matter if the slider was reversed or not. By default the arrow keys are oriented by arrow up/right to the higher slider value, arrow down/left to the lower slider value. |
 | ticks | array | [] | Used to define the values of ticks. Tick marks are indicators to denote special values in the range. This option overwrites min and max options. |
-| tick_labels | array | [] | Defines the labels below the tick marks. Accepts HTML input. |
-| snap | float | 0 | Used to define the bounds of a tick. Snaps to the tick if value is within these bounds. |
+| ticks_labels | array | [] | Defines the labels below the tick marks. Accepts HTML input. |
+| ticks_snap | float | 0 | Used to define the snap bounds of a tick. Snaps to the tick if value is within these bounds. |
 
 Functions
 =========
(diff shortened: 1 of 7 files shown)


seiyria:
Very nice! Could you please provide a jsfiddle so we can check this out, and possibly some tests?

MelleB:
Thanks, code looked clean and was easy to read! 
I created an example in the index.tpl file (at the bottom, no. 12).
What would you like to see in the jsfiddle/tests?

seiyria:
Regarding a jsfiddle, it would just be nice to see what it looks like w/o having to clone and run.

As for tests, here's what I can think of:

* Verify that the number of tick marks matches what you set
* Verify the tick marks are at the correct intervals

MelleB:
jsfiddle: http://jsfiddle.net/qn3u9orj/

seiyria:
This looks nice. A bit different than I originally expected (my implementation was different), but I think this is good. @rovolution?

rovolution:
@MelleB @seiyria I would like to see automated tests for a feature of this magnitude

README.md
```diff
@@ -125,6 +125,8 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | enabled | bool | true | whether or not the slider is initially enabled |
 | formatter |	function |	returns the plain value |	formatter callback. Return the value wanted to be displayed in the tooltip |
 | natural_arrow_keys | bool | false | The natural order is used for the arrow keys. Arrow up select the upper slider value for vertical sliders, arrow right the righter slider value for a horizontal slider - no matter if the slider was reversed or not. By default the arrow keys are oriented by arrow up/right to the higher slider value, arrow down/left to the lower slider value. |
+| ticks | array | [] | Used to define the values of the ticks. Overwrites min and max options. |
```
rovolution: @MelleB  I would like this to be more descriptive as to what a "tick" actually is

MelleB:
@rovolution Thanks for the feedback, added tests and extended the ticks option description a bit.

MelleB:
JSFiddle with tick mark labels: http://jsfiddle.net/qn3u9orj/1/

less/bootstrap-slider.less
```diff
@@ -151,3 +162,31 @@
 		}
 	}
 }
+.slider-tick {
+	position: absolute;
+	width: @slider-line-height;
+	height: @slider-line-height;
+	#gradient > .vertical(#f9f9f9, #f5f5f5);
+	.box-shadow(inset 0 -1px 0 rgba(0,0,0,.15));
+	//.box-shadow(~"inset 0 1px 0 rgba(255,255,255,.2), 0 1px 2px rgba(0,0,0,.05)");
```
rovolution: delete commented code

less/bootstrap-slider.less
```diff
@@ -151,3 +162,31 @@
 		}
 	}
 }
+.slider-tick {
+	position: absolute;
+	width: @slider-line-height;
+	height: @slider-line-height;
+	#gradient > .vertical(#f9f9f9, #f5f5f5);
```
rovolution: Is this #gradient selector necessary? Also want to refrain from using IDs in stylesheets.

rovolution:
@MelleB Just looked through this PR. Awesome work here!! Just left some minor feedback/comments.

Besides the feedback I left inline, one more thing I would like to see a namespace for the tick related arguments. Something like the following would be good:

ticks
ticks_labels
ticks_snap

Once you get that in I'll review and merge it in!

MelleB:
@rovolution Thanks for the review. I renamed the options per your suggestion and removed the commented code. The `#gradient` part is required as far as i can tell because otherwise the rendering of the ticks is missing a background.

rovolution:
@MelleB just pulled down your branch and played around with it.

The arrow key functionality appears to be broken for sliders with ticks.

Also, is there any way we can get the slider handle to auto-snap to the nearest tick?

MelleB:
@rovolution Thanks for testing this. An unfortunate side effect I overlooked.

What happens is that any value within the tick's bounds gets set to the tick value. Since keyboard navigation uses the standard step option (1 by default) which is smaller than the snap value of 30 in the example, the value gets reset to 0.

I think the best thing to do here is to ignore the snapping when using keyboard navigation. Another option would be to introduce extra navigation options (e.g. CTRL+Arrow = move to tick) but I don't think many users will notice/use this feature.

What are your thoughts on this?

MelleB:
Fixed + I needed an 'in-selection' class on the ticks.

rovolution:
@MelleB awesome! could you add an automated test to your spec that tests if the keyboard functionality still works when ticks are enabled?

That should be it! Our ```KeyboardSupportSpec.js``` file has an example of how to simulate keyboard events

MelleB:
Seems Travis CI is failing because the connection to github.com times out...

seiyria:
There's supposed to be an option to re-run the build, but I can't seem to find it.

seiyria:
I'll monitor it and see if I can get it to evaluate properly.

MelleB:
:+1

seiyria:
It passed this time around.

rovolution:
@MelleB merged. Thanks for taking the time to put together a great PR!

rovolution:
@MelleB I just noticed that the gradient track style (the faint blue) now gets applied to all sliders, regardless of whether they specified ticks or not.

Could you scope those styles to some class tag to ensure they only appear when the user has specified ticks?

MelleB:
Please see #304

Issue #199, opened 2014-09-05
Range selector, 3 color


KANekT:
![future](http://cdn.joxi.ru/uploads/prod/2014/09/05/c36/3ce/577d43eadca7db83965c48f7fc53c8c0c165c293.jpg)

I's possible ?

will-e-yams:
implemented in v 4.5.6. close this?

PR #297, opened 2015-01-30
Add left/top and right/bottom track segments so they can be styled independently

mgwalker:
This adds left and right track segments that can be targeted by CSS (with .slider-track-left and .slider-track-right, respectively).  It allows users to style the "unselected" part of the slider, which is useful if you're building a "red-amber-green" style slider.

jsfiddle: http://jsfiddle.net/qfjjv2vp/

### DIFF
--- a/sass/slider.scss
+++ b/sass/slider.scss
@@ -41,7 +41,7 @@ $baseBorderRadius: 4px;
       top:  50%;
       left: 0;
     }
-    .slider-selection {
+    .slider-selection, .slider-track-left, .slider-track-right {
       height: 100%;
       top: 0;
       bottom: 0;
@@ -74,6 +74,11 @@ $baseBorderRadius: 4px;
       top: 0;
       bottom: 0;
     }
+	.slider-track-left, .slider-track-right {
+	  width: 100%;
+	  left: 0;
+	  right: 0;
+	}
     .slider-handle {
       margin-left: -$baseLineHeight/4;
       margin-top: -$baseLineHeight/2;
@@ -108,6 +113,13 @@ $baseBorderRadius: 4px;
   	@include box-shadow(inset 0 -1px 0 rgba(0,0,0,0.15));
   	@include box-sizing(border-box);
 	@include border-radius($baseBorderRadius);
+}
+
+.slider-track-left, .slider-track-right {
+  position: absolute;
+  background: transparent;
+  .box-sizing(border-box);
+  border-radius: border-radius($baseBorderRadius);
 }
 
 .slider-handle {
--- a/css/bootstrap-slider.css
+++ b/css/bootstrap-slider.css
@@ -39,7 +39,9 @@
   top: 50%;
   left: 0;
 }
-.slider.slider-horizontal .slider-selection {
+.slider.slider-horizontal .slider-selection,
+.slider.slider-horizontal .slider-track-left,
+.slider.slider-horizontal .slider-track-right {
   height: 100%;
   top: 0;
   bottom: 0;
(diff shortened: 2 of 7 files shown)


seiyria:
This looks pretty nice, I gotta say. Would you make some tests for situations where you color each of these components, as well as verify the defaults are correct?

mgwalker:
I'll be glad to make some tests, but I'm honestly not sure how to go about writing tests for this.  It seems like it'll be a manual thing since you've actually got to look at the rendered slider to verify it's correct.  Do you just want a sample added?

I promise I'm not trying to be a pain, I'm just not sure how to give you what you're asking for.  :)

seiyria:
Sure, no worries. Basically, create the slider as you have in your tests, and just verify that, after creation, the respective slider tracks actually have their colors set appropriately (select them, check their color).

mgwalker:
So, here's what I'm guessing:  I add input elements to SpecRunner.tpl, then add spec that creates the sliders, selects them, and checks the colors?  Since I'll need CSS to style them, is there somewhere I can put a CSS file for testing (I assume you don't want it inlined in the SpecRunner.tpl)?

seiyria:
You can probably just inline it; I believe that's how the other slider CSS is handled. But yeah, that's the gist of it. That's also how the example page is done.

mgwalker:
I added some tests that verify the background color as well as the left and right segment widths.  Hopefully this is what you had in mind.  :)

seiyria:
Hey, it looks great to me. @rovolution - what do you think?

mgwalker:
Incidentally, I just found #199.  This could satisfy @KANekT's needs as well, so that's cool.

rovolution:
@mgwalker @seiyria just finished reviewing. This is a really well-done PR...thanks!

One more thing before I merge this in...could you add an example to the ```/tpl/index.tpl``` file?

It can be something simple where it shows how you can use CSS to style different parts of the track

mgwalker:
Sure thing, I'll do it first thing Monday morning.

mgwalker:
Added a sample to index.tpl.  Let me know if there's anything else I should do!

rovolution:
@mgwalker one final thing...could you add those styles to our .sass file as well?

https://github.com/seiyria/bootstrap-slider/blob/master/sass/slider.scss

other than that, this looks great!

mgwalker:
Sure.  I'm curious why you have both SASS and LESS, though.  Seems like extra effort to maintain, but no gain.  That's a discussion for another place, though.

seiyria:
Same reason tools like Bootstrap do it; it lets the users use whichever tool they're most comfortable with go down that route.

mgwalker:
Fair enough.  I've updated the SASS to match the LESS.

rovolution:
@mgwalker merging in! thanks for the contribution!

sass/slider.scss
```diff
@@ -110,6 +115,13 @@ $baseBorderRadius: 4px;
 	@include border-radius($baseBorderRadius);
 }
 
+.slider-track-left, .slider-track-right {
+  position: absolute;
+  background: transparent;
+  .box-sizing(border-box);
```
jamesgary: We're using grunt and sass, and this line gives us the following error when including this file:

```
Line 121 of node_modules/bootstrap-slider/sass/slider.scss: Invalid CSS after "  .box-sizing": expected "{", was "(border-box);"
```

Is this valid sass?

mgwalker: Apparently not.  It should probably be

```@include box-sizing(border-box);```

I've never used SASS before and it's not in the default Grunt task, so I probably never ran it.

rovolution:
@jamesgary just hotfixed

jamesgary:
Thanks! :two_hearts:

rovolution:
@mgwalker As I was thinking about this, I thought make more sense to call these class tags ```slider-track-low``` and ```slider-track-high``` versus left and right. The reason being, lets say you have a situation where the user has a vertically-aligned slider. The value range makes more sense versus the direction, which always depends on the orientation.

If its not too much trouble, would you be able to make those changes?

mgwalker:
Yes, I've create a new pull request in #305 for the changes.

PR #300, opened 2015-02-03
Add support for tooltip_separator option.

tvararu:
Despite the fact that this option is mentioned in the documentation, I could not find it being implemented in any way in the codebase, so I did.

The implementation is poor as it changes the method signature for `formatter`. I did however write a basic test which should pass.

Let me know what else I should change, and I'll update my PR accordingly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -371,7 +371,7 @@ $("#ex5").slider();
 var slider = new Slider('#ex5');
 
 $("#destroyEx5Slider").click(function() {
-	
+
 	// With JQuery
 	$("#ex5").slider('destroy');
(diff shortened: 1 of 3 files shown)


rovolution:
@tvararu we made the decision to remove this feature and apparently forgot to remove it from our docs!

The reasoning being was that it represented feature creep with the ```formatter``` method already in place

tvararu:
Ah, fair enough. I'll submit a PR to update the doc, then?

rovolution:
@tvararu yea that would be perfect! thanks!

PR #302, opened 2015-02-05
(feat) Add logarithmic scale

MelleB:
This commit allows the scale to be changed from `linear` to
`logarithmic` using the newly introduced `scale` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | ticks | array | [ ] | Used to define the values of ticks. Tick marks are indicators to denote special values in the range. This option overwrites min and max options. |
 | ticks_labels | array | [ ] | Defines the labels below the tick marks. Accepts HTML input. |
 | ticks_snap_bounds | float | 0 | Used to define the snap bounds of a tick. Snaps to the tick if value is within these bounds. |
+| scale | string | 'linear' | Set to 'logarithmic' to use a logarithmic scale. |
 
 Functions
 =========
--- a/index.html
+++ b/index.html
@@ -714,6 +714,41 @@ var slider = new Slider("#ex13", {
   </div>
 
 
+      	<div class='slider-example'>
+      		<h3>Example 14:</h3>
+      		<p>With a logarithmic scale.</p>
+      		<div class="well">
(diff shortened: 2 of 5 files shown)


MelleB:
I'll post a JSFiddle in a minute

seiyria:
Don't forget the tests!

MelleB:
Minimal test and an example are in there. Not sure how to go about keyboard navigation in relation to the logarithmic values. Do we want to keep using step here?

seiyria:
I'm actually not sure what would be most natural, to be honest.

MelleB:
JSFiddle: http://jsfiddle.net/v8htqnfz/

MelleB:
Rebased branch

rovolution:
@MelleB looks like the keyboard functionality is broken on the logarithmic slider

MelleB:
@seiyria  @rovolution Yes, I didn't touch any of the keyboard stuff yet. I'm not sure how keyboard navigation on a logarithmic slider should look like. Would we just use the default (linear) step here?

AFAICT using step size results in extremely long key-presses. Using the logarithmic increments on the other hand results in loss of precision. (And how would you configure the step option if we do use logarithmic increments?)

Would love some input on this.

seiyria:
Honestly I'm quite unsure about this. I think it should probably use step size, and if someone has problems, they can slide it another way. I was hoping to see what [this library](http://refreshless.com/nouislider/slider-values/) does, but it seems like they don't even have keyboard support.

rovolution:
I would go ahead and use step size, despite the fact it may lead to long key presses. If people leave feedback we can always incorporate it later

rovolution:
@MelleB The keyboard functionality for the range slider appears to be broken.

Its throwing the following error:

```
Uncaught TypeError: Cannot read property 'toFixed' of undefined
```

Could you fix that? Im looking through our KeyboardSupportSpec and am somewhat confused how our tests are not breaking on that

MelleB:
Didn't see that error, weird. What browser are you using?

Sent from mobile
Op 14 feb. 2015 19:04 schreef "Rohit Kalkur" <notifications@github.com>:

MelleB:
Looking into it...

MelleB:
This fixes the issue.

rovolution:
@MelleB merged! Thanks again for this awesome PR

PR #304, opened 2015-02-06
(fix) Track selection for ticks gets an extra selector

MelleB:
In the previous pull request #296  @rovolution noted I took the liberty to change the default styling of the track selection for all sliders. This fixes that issue :)


rovolution:
@MelleB if you could modify the SASS file as well that would be perfect! Thanks again

rovolution:
@MelleB merged. Thanks!

PR #309, opened 2015-02-12
Calculated offsets incorrect when body has a margin of auto

rheidari:
Calculated slider offsets are incorrect when the body has a margin of auto - resulting in an offset in display when you initiate the drag and where the handle ends up being displayed. 

I've changed the do..while loop with offsetParent to use getBoundingClientRect() and it appears to fix the issue.

Broken Example: http://jsfiddle.net/0rh54yxo/1/
Working Example (using getBoundingClientRect): http://jsfiddle.net/0rh54yxo/6/

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1298,9 +1298,9 @@
 				element.className = newClasses.trim();
 			},
       _offset: function (obj) {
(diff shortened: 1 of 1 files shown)


rovolution:
@rheidari merged. thanks for taking the time to fix that!

Issue #317, opened 2015-02-24
[Blocker] Calling "setValue" with array may fail


crevete:
When I call ```setValue``` method on a range slider, it fails with the following error :
```
Invalid input value '5,100' passed in
```
YES, please not that ```[] instanceof Array``` may return ```false```. Cf. http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/

I suggest using the native API ```Array.isArray([])``` or jQuery API ```$.isArray([])``` for better compatibility, rather than ```instanceof Array```.

Could you get it fixed ASAP, thanks

seiyria:
Please make a jsfiddle so we can actually see the issue.

crevete:
It's difficult... because my application runs with GWT. I does not use directly your js file, but a GWT wrapper. I've tested the type of the first argument with ```instanceof Array``` just before passing it to the ```setValue``` function, so I'm sure that it's an Array. But as described in http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/ : 
```
The problems arise when it comes to scripting in multi-frame DOM environments. In a nutshell, Array objects created within one iframe do not share [[Prototype]]’s with arrays created within another iframe. Their constructors are different objects and so both instanceof and constructor checks fail:

var iframe = document.createElement('iframe');
document.body.appendChild(iframe);
xArray = window.frames[window.frames.length-1].Array;
var arr = new xArray(1,2,3); // [1,2,3]

// Boom!
arr instanceof Array; // false

// Boom!
arr.constructor === Array; // false
```

That's why I've created this issue. In fact, just to debug it, I've changed ```instanceof Array``` to ```$.isArray()``` in the JS source code, and this have get it fixed.  So I just suggest using a better function to detect if a variable is an array.

crevete:
Can you fix this please? Otherwise I have to patch the JS file...

seiyria:
Feel free to submit a PR; both of us do have other projects on our plate and this one is mostly run by the community that uses it.

crevete:
Thanks

PR #320, opened 2015-02-26
Use "Array.isArray" to check whether an object is an array

crevete:
Fix for #317

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -345,7 +345,7 @@
 
 				/* Create ticks */
 				this.ticks = [];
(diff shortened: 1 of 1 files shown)


seiyria:
Seems fine. @rovolution?

rovolution:
@seiyria looks good to me. Found this good StackOverflow post detailing the differences between ```Array.isArray``` vs ```instanceof Array```

http://stackoverflow.com/questions/22289727/difference-between-using-array-isarray-and-instanceof-array

seiyria:
I see. That string check, while compatible, is pretty ugly. I think this will be fine since we only really support evergreen browsers anyway. Thanks for sharing the link.

Will merge.

PR #325, opened 2015-03-02
Fixed getting options with complicated names from data attributes

kirilvit:
Changed underscores with dashes for correct getting data attributes values

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -288,7 +288,7 @@
 			}
 
 			function getDataAttrib(element, optName) {
-				var dataName = "data-slider-" + optName;
+				var dataName = "data-slider-" + optName.replace(/_/g, '-');
 				var dataValString = element.getAttribute(dataName);
 
 				try {
--- a/test/specs/ElementDataAttributesSpec.js
+++ b/test/specs/ElementDataAttributesSpec.js
@@ -45,6 +45,24 @@ describe("Element Data Attributes Tests", function() {
     var sliderValue = slider.slider('getValue');
     expect(sliderValue).toBe(5);
   });
+  
+  it("reads the 'data-slider-ticks-labels' property and sets it on slider", function() {
+    slider = $("#sliderWithTickMarksAndLabels").slider();
+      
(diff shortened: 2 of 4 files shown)


seiyria:
Seems simple enough. Do we need tests for this kind of behavior @rovolution?

rovolution:
@seiyria I would not think so (I think the current tests we have in the DataAttributesSpec file should cover this), but it would be nice to find out which specific data attribute was the cause of this change, and at least add a test for that to the DataAttributesSpec file. Which data attribute caused you to make this change @kirilvit?

kirilvit:
@rovolution it was `ticks_labels`. If you try to set it via data attribute, it would not work.

Actually, it could be any attribute with underscore in name.

rovolution:
@kirilvit cool. yea if you could add a test to the [ElementDataAttributesSpec](https://github.com/seiyria/bootstrap-slider/blob/master/test/specs/ElementDataAttributesSpec.js) file for the ```tick_labels``` attribute, that would be perfect.

kirilvit:
Ok,I'll do it tomorrow

rovolution:
@kirilvit awesome thanks!

kirilvit:
How do you run tests? I'm new to Jasmine, and when I run Grunt to build SpecRunner.html from tpl it throws an error  
`Running "jasmine:src" (jasmine) task
Warning: Error while processing "tpl/SpecRunner.tpl" file. Used --force, continuing.`

rovolution:
should be just ```grunt test```

kirilvit:
Thanks. I used a different version of Jasmine, and that's why it was failing.

kirilvit:
I've just committed a test for this fix. And also I've updated examples page, replaced `data-slider-ticks-labels="['$0', '$100', '$200', '$300', '$400']"` with `data-slider-ticks-labels='["$0", "$100", "$200", "$300", "$400"]'`, because JSON.parse needs correct JSON.

rovolution:
This looks good...merging in. Thanks for the PR @kirilvit!

PR #326, opened 2015-03-02
Fixed issue with touch-enabled laptops using a mouse

### LINKED ISSUE #293: Chrome Version 40.0.2214.91m - Slider broken
Hi,

In the latest stable Chrome I find the slider is broken completely for mouse.  I can use keyboard after I've clicked on the thumb, but mouse moving doesn't work.

I took a guess that this was due to chrome's handling of touch events and it's idea of mouse capture.  As a real dirty fix for me (and so I could patch our production quickly) I set this.touchCapable = false; on line 336 of bootstrap-slider.js

I appreciate this isn't a good fix, and I'm not really that knowledgeable with Chrome's touch API to offer a better solution.  But if anyone else has been having issues today with the slider not working in Chrome, figured this might help.

Alex

### LINKED ISSUE #312: touch-events chrome flag breaks draggablity
Version 40.0.2214.95 (64-bit)
Linux Mint 17.1 
chrome://flags/#touch-events -> Enabled   results in immovable slider handle, can't be dragged with mouse, touchpad, or touchscreen.  

Searched through related issues but seems to be slightly different than existing issues.  This issue is not a result of modernizr as removing it does not fix.

JeroenWarmerdam:
This change fixes issues that appear when a user has a touch-enabled computer. The current implementation detects that the system is touch capable, but does not take into account that the user might be using a regular mouse instead of the touch screen.

fixes #293 
fixes #312

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -583,11 +583,11 @@
 				// Bind touch handlers
 				this.mousedown = this._mousedown.bind(this);
 				this.sliderElem.addEventListener("touchstart", this.mousedown, false);
-			} else {
-				// Bind mouse handlers
-				this.mousedown = this._mousedown.bind(this);
-				this.sliderElem.addEventListener("mousedown", this.mousedown, false);
 			}
+            // Bind mouse handlers
+            this.mousedown = this._mousedown.bind(this);
+            this.sliderElem.addEventListener("mousedown", this.mousedown, false);
+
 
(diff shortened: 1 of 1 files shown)


seiyria:
Can you please submit a jsfiddle and make your code pass Travis?

rovolution:
We can't accept this until the tests are passing @JeroenWarmerdam

rovolution:
fixed this and got it to pass the tests. going to close this

Issue #330, opened 2015-03-11
Option to focus slider handle after click or drag


riga:
After changing the slider value (either via click or drag) it would be nice
to focus the slider handle so the user can immediately use arrow keys to
change the value again.

I think this could be a configurable option.

seiyria:
Seems like a nice idea.

riga:
Fixed in PR #332.

PR #332, opened 2015-03-12
Add focus option.

riga:
I think the way ``_triggerFocusOnHandle()`` was used before (in ``_mousedown()``),
i.e. without any arguments, was unintended in the first place. Now, it is called with
the appropriate slider number after the value change took place if the ``focus``
option is ``true``.

Also updated docs and index template.
I don't think that a unit test is required here.

When this PR is accepted, #330 is fixed and can be closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | ticks_labels | array | [ ] | Defines the labels below the tick marks. Accepts HTML input. |
 | ticks_snap_bounds | float | 0 | Used to define the snap bounds of a tick. Snaps to the tick if value is within these bounds. |
 | scale | string | 'linear' | Set to 'logarithmic' to use a logarithmic scale. |
+| focus | bool | false | Focus the appropriate slider handle after a value change. |
 
 Functions
 =========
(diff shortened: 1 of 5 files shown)


seiyria:
Please provide tests and a jsfiddle to supplement our PR. thanks!

riga:
Here is the jsfiddle: http://jsfiddle.net/vqpL0x93/.

seiyria:
Looks nice. I'm also unsure if this really has to be an option, I think this would be fine as default behavior. I'll wait on @rovolution to make a decision on that though.

rovolution:
I like having it as an option. I know we have had people in the past complain if the focus is only on the handle. I would really like some sort of unit test on this though

riga:
Is the unit test ok?

seiyria:
I think it looks right.

rovolution:
@riga looks good. could you squash your commits?

rovolution:
@riga dont worry about the squash this looks good im going to merge in

rovolution:
@riga btw thanks for the PR!

riga:
This is a really nice project! I'm glad I could help ;)

FYI: We use bootstrap-slider in a physics related webapp that helps us manage
large amounts of histograms (~10k) created for the search of the Higgs boson at CERN.

seiyria:
Awesome! Glad to hear it's being used in some awesome apps in production!

Issue #327, opened 2015-03-07
Slider not working inside bootstrap comun grid columns


sades:
Hi, I am trying to use this slider but when its placed inside any div that has a class="col-md-6 for example it does not work correctly none of the col-*-* are working when placed inside those divs and you try to slide it just goes to the end of the slide.

seiyria:
Did you add the col-md-12 class to the slider itself?

sades:
no i even try reproducing this http://jsfiddle.net/2DygG/7/ it seems to work fine in the fiddle because the limits of the iframe are small but out of it it wont work

seiyria:
It's not because of the iframe, it's because of this:

```
<input style="width: 100%;" />
```

Did you try this?

sades:
yes i did if you open that fiddle http://jsfiddle.net/2DygG/7/  and do this 
![01](https://cloud.githubusercontent.com/assets/271004/6539721/f2736324-c449-11e4-9233-aae9652dcc3e.jpg)
Just click on it and it will go directly to the end to move it you basically have to click over it and hold the mouse and move it to the end of the screen

seiyria:
Strange, but I can repro this now at least.

sades:
Im on a resolution of 1920x1080 i changed to 1024x762 and it worked almost perfectly seems it doesn't like big screens or resolutions, tested on my 15 inch labtop also and works better than on the 22 inch screen but still you have to move way more to the borders of the screen

sades:
Ok doing some more test I decided to place direct styles to the div that contains the slider but simulating bootstrap styles, what I encounter is that the position relative that have bootstrap divs cause the marker to function incorrectly

Example
<div style="width:90%; float:left; min-height: 1px;  padding-left: 15px; padding-right: 15px; position: relative;">

When i test the same div but without the position relative like this:

<div style="width:90%; float:left; min-height: 1px;  padding-left: 15px; padding-right: 15px;">

It works like a charm, hope this helps to get a fix around

shammar13:
@sades, I've got the same issue. I have also found that setting the position of the divs that contain the slider to be static fixes the issue.

Maybe there is an error in how the position of the mouse/click is calculated when the slider is in a relative div?

@seiyria, I found another way for you to reproduce the issue. Go to your examples page (http://seiyria.github.io/bootstrap-slider/) and then inspect element on one of the `.well` elements and then just add the style `position: relative;` and then try dragging the slider around.

[...]

PR #334, opened 2015-03-14
Fix bug with slider within Bootstrap columns

mark47:
Fix to bug where slider would move to max (and then freeze there) when using within a Bootstrap column with a large width. Related to Bootstrap’s use of `position: relative`

Discussed in [issue #327](https://github.com/seiyria/bootstrap-slider/issues/327)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bootstrap-slider",
-  "version": "4.5.3",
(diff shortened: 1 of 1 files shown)


seiyria:
Please provide a jsfiddle to show these changes, and if you could, please rebase out the latter two commits, as they seem wholly pointless.

PR #335, opened 2015-03-14
Fix bug with slider within columns and modal

### LINKED ISSUE #327: Slider not working inside bootstrap comun grid columns

crevete:
fix #327

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1337,12 +1337,15 @@
 
 				element.className = newClasses.trim();
 			},
-		    _offset: function (obj) {
-		        return {
-		          left: obj.offsetLeft,
-		          top: obj.offsetTop
-		   		};
-		    },
+			_offset: function (obj) {
+				var rect = obj.getBoundingClientRect(),
+				ol = rect.left,
+				ot = rect.top;
+				return {
+					left: ol,
+					top: ot
+				};
+			},
 			_css: function(elementRef, styleName, value) {
                 if ($) {
                     $.style(elementRef, styleName, value);


seiyria:
Please provide a jsfiddle demo showing that this is fixed.

crevete:
@seiyria you don't read messages ? I've put a link in the issue #327...
http://jsfiddle.net/2DygG/11/

crevete:
And please take care when you release a new version. FYI, the 4.5.4 tag contains a wrong version of js file AGAIN...

seiyria:
@crevete:

Please read our CONTRIBUTING.md -- it's not there to be ignored. Regardless of whether or not you've posted a link in another thread, it should be here too _so all PRs are tied to a demo_. Always.

Also, please be a bit more respectful. Just as you offer your free time to submit a PR to us, we offer our free time to review/merge PRs and manage this project, and it's not easy.

@rovolution, what do you think about the PR?

crevete:
@seiyria I've never read your CONTRIBUTING.md, sorry about this and I take note now. But on your side, it's not the first time that a new tag that fixes some bugs but introduces some ones ; or it contains wrong version of files. It's really not cool. I know it's not easy, so I just ask you to take care when you review PR requests or release a tag. Thanks!

seiyria:
If you don't like regressions, feel free to submit more tests to help prevent regressions in the future. That would be the best thing you could do. We can't test every possibility that our users will be engaging in; that's simply not feasible.

crevete:
I don't ask you to test every possibility, and I know regressions cannot be avoid. Just take care to minimize the risk. If a regression is introduced, especially a blocker one like this (for some users off course), please be more reactive (if you can) to fix it, rather than wait for people blocked by the issue to analyze and find the fix. Because you are responsible of this project, and maybe many people are waiting for the fix. Anyway, if you don't have time, forget what I say.

But for the packaging issue, I think you guys are able to do better. Thanks!

rovolution:
@crevete apologizes for this. That code was modified to fix vertical sliders, which were broken by those very lines you added in.

@seiyria and I are both very busy people during the weekday and we try our best to fix things as quickly as possible. However, we both are full time working professionals who have other obligations beyond this project. I had a very long/exhausting  at the office and had little to no time to address slider issues.

Thank you for your patience. I will add this back in for now, and we will begin investigating a better long-term solution for fixing vertical sliders for good

crevete:
@rovolution thank you for your reply :+1: So #327 may be closed now.

PR #354, opened 2015-04-07
Honor step size on the logarithmic scale.

manthey:
Added a test to make sure the step size was honored.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -231,7 +231,17 @@
 				toValue: function(percentage) {
 					var min = (this.options.min === 0) ? 0 : Math.log(this.options.min);
 					var max = Math.log(this.options.max);
-					return Math.exp(min + (max - min) * percentage / 100);
+					var value = Math.exp(min + (max - min) * percentage / 100);
+					value = this.options.min + Math.round((value - this.options.min) / this.options.step) * this.options.step;
+					/* Rounding to the nearest step could exceed the min or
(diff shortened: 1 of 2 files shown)


seiyria:
Seems fine. I'm not sure if this needs to be documented since it should be expected behavior anyway. @rovolution?

js/bootstrap-slider.js
```diff
@@ -231,7 +231,17 @@
 				toValue: function(percentage) {
 					var min = (this.options.min === 0) ? 0 : Math.log(this.options.min);
 					var max = Math.log(this.options.max);
-					return Math.exp(min + (max - min) * percentage / 100);
+					var value = Math.exp(min + (max - min) * percentage / 100);
+					value = this.options.min + Math.round((value - this.options.min) / this.options.step) * this.options.step;
```
rovolution: you are initializing `value` on line 234 only to re-set its value on the following line

rovolution:
@manthey thanks for the PR! Could you see my comment on the code, make the update, and tag me again?

Thanks!

js/bootstrap-slider.js
manthey: @rovolution Line 234 calculates the unstepped value.  Line 235 uses the value calculated in line 234 and rounds it to a stepped value.  It seemed clearer to have it in two lines than in one.

rovolution: oh i see it now! thanks for clarifying that!

Issue #356, opened 2015-04-09
Bug generating wrong 'in-selection' ticks when selection === 'after'


TheDartCode:
When you set selection:'after' and use ticks, the wrong ticks are marked as 'in-selection'.

Fiddle: https://jsfiddle.net/zbc3gs1L/

seiyria:
Which ones would you expect to be marked `in-selection`? It seems like everything after the cursor is being marked, which seems accurate.

TheDartCode:
I would expect this: https://jsfiddle.net/en7wuh4L/

I have implemented the changes in this commit,
https://github.com/TheDartCode/bootstrap-slider/commit/e618a6a46d0266087fe316b72e133e1705f82660
I'll issue a pull request if you accept.

seiyria:
Ahhhhh. I see how it works now. Yes, I think it's fine to submit that as a PR if you include a test as well. Thanks!

TheDartCode:
Hehe, will do asap :)
Thank you for the great component!

seiyria:
And thank you for using it and helping improve it!

PR #357, opened 2015-04-09
Fix issue #356

TheDartCode:
Bug producing wrong 'in-selection' ticks


seiyria:
Very nice. @rovolution?

PR #359, opened 2015-04-10
Fix offset calculation when contained in relatively positioned parent. Fixes #340

### LINKED ISSUE #340: Using bootstrap-slider as dropdown menu[bug]
Everything is ok, except using it as a dropdown menu.When I move the slider after dropdown, it is not the exact position where I press.
http://stackoverflow.com/questions/29142031/using-bootstrap-slider-as-dropdown-menu

bcroesch:
Apologize if all my pushes were spamming anyone. Kept trying to get the damn white space to line up (looked fine in my editor).

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1340,23 +1340,25 @@
 
 				element.className = newClasses.trim();
 			},
+			_offsetLeft: function(obj){
+				var offsetLeft = obj.offsetLeft;
+				while((obj = obj.offsetParent) && !isNaN(obj.offsetLeft)){
+					offsetLeft += obj.offsetLeft;
+				}
+				return offsetLeft;
+			},
+			_offsetTop: function(obj){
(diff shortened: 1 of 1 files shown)


seiyria:
Nope, no one got spammed by this. Thanks. Can we see this in action via a jsfiddle or some equivalent?

bcroesch:
Original fiddle from the issue (shows working & broken versions): http://jsfiddle.net/KyleMit/dy0qobyp/4/

Fiddle with fix (both working): http://jsfiddle.net/gd92637g/1/

rovolution:
how does this look inside of a modal? we had people complain about the slider when it was contained within a modal

could you add that to your PR?

bcroesch:
Here you go: http://jsfiddle.net/gd92637g/2/

seiyria:
Seems like it works to me. Thanks!

Issue #364, opened 2015-04-20
Overlapping handles do not swap reliably due to precision of percentage comparison


thomato:
Hi there,

I noticed the following when using the slider.

In the following code
```javascript
			_adjustPercentageForRangeSliders: function(percentage) {
				if (this.options.range) {
					if (this.dragged === 0 && this.percentage[1] < percentage) {
						this.percentage[0] = this.percentage[1];
						this.dragged = 1;
					} else if (this.dragged === 1 && this.percentage[0] > percentage) {
						this.percentage[1] = this.percentage[0];
						this.dragged = 0;
					}
				}
			},
```
Overlapping handles do not swap reliably due to precision of percentage values during comparison.

In a particular case,
values in this.percentage are 45.45454545454545 while values in percentage is 45.45454545454546

I prepared the case in the fiddle https://jsfiddle.net/ewsme/6xsag3jy/

seiyria:
I honestly don't know what we can do about this. I believe we give you an option to control the precision, so perhaps round it a bit?

thomato:
I have a proposed fix here http://jsfiddle.net/ewsme/sogjrkby/

thomato:
Was about to do a pull request

seiyria:
Hm. If you have a fix then there's no harm in submitting it. I just am not sure offhand what you can do about floating point precision besides arbitrarily capping it (which I'm not sure we want to do), but I don't know what you came up with.

PR #365, opened 2015-04-20
Fix for issue #364

thomato:
Here is the proposed fix by capping the precision according to the percentage.


thomato:
Suspect an issue with the fix. Will review before resubmitting.

PR #366, opened 2015-04-20
Fix for issue #364

thomato:
Truncate percentage for comparison based on percentage to be compared


seiyria:
Hm, this seems ok I think. @rovolution? Maybe a test would be nice but I'm not sure here.

rovolution:
I would like a test of some sort...if anything to validate the slider value

thomato:
Got it. Working on that. Thanks for the feedback.

thomato:
Added test for handle swapping. Would appreciate any feedback.

seiyria:
It seems fine, although I'm not sure you needed to bump it. We have a grunt task for that.

thomato:
I think the bump was from the merge, didn't bump for my changes.

seiyria:
Then sure, it looks fine to me. @rovolution?

rovolution:
is there anyway we could write the tests to call either the `setValue` method or simulate a `mousemove` event like [this test](https://github.com/seiyria/bootstrap-slider/blob/master/test/specs/ConflictingOptionsSpec.js#L16-L31)?

I would rather we simulate the value change that way, since that is our "public" interface, and if the value is what we expect from using our public method, then we can assume the private method is implemented properly.

thomato:
@rovolution Thanks for the tip. Updated the tests to use mouse events instead.

thomato:
hello, just wanted to inform that the test is ready, not sure if anyone had the chance to look at this yet.

seiyria:
Not yet. To be honest, it looks ok to me but I'm not used to tests that use the mouse yet.

thomato:
OK, no worries, just checking.

rovolution:
@thomato sorry for the delay! I am on vacation right now and will merge/bump the version when I get back

thomato:
No problem, thank you.

thomato:
Thanks for the merge!

rovolution:
thank you for the contribution!

PR #367, opened 2015-04-25
prevent overlay of unused and invisible tooltip element

bafio:


seiyria:
please provide a jsfiddle so we can see your changes, and if you could please describe what you're changing (possibly with reference to related issues where applicable), that would be great!

bafio:
https://jsfiddle.net/fabiotrabucchi/crb1oo30/ --> in this case I set the slider with
```
{tooltip: 'always', tooltip_split: false}
```
and I've altered the css
```
.tooltip.tooltip-max, .tooltip.tooltip-min { opacity: 0.3 }
```
to show the issue (.tooltip.tooltip-max becomes .tooltip.tooltip-max.bottom and goes on top of the buttons below, preventing their the "click" because something else is on top (and it is not visible because the default opacity is 0))

https://jsfiddle.net/fabiotrabucchi/5nkdqskx/ --> as before I set the slider with
```
{tooltip: 'always', tooltip_split: false}
```
and I've altered the css
```
.tooltip.tooltip-max, .tooltip.tooltip-min { opacity: 0.3 }
```
to show that the issue is gone since ```.tooltip.tooltip-max, .tooltip.tooltip-min { display: none }``` when ```{tooltip_split: false}```, therefore if .tooltip.tooltip-max(or min) becomes .tooltip.tooltip-max(or min).bottom(or top) nothing goes over other elements

seiyria:
Alternatively, what if you were to simply do `pointer-events: none;` on the tooltips? Not that that's a better solution, but I'm curious of the differences in the two approaches.

bafio:
It might be a good option, not well supported on IE as the display property http://caniuse.com/#feat=pointer-events

seiyria:
Ah, drat, that would break what browser support we claim to have, too. Thanks.

seiyria:
@rovolution - what do you think about this change?

rovolution:
this looks good to me.

@bafio could you add an automated test that verifies the fix?

So if `tooltip_split: false`, then `tooltip_min` and `tooltip_max` should have `display: none` and `this.tooltip` should have the `in` class applied

Otherwise (`tooltip_split: true`), then `this.tooltip` should have `display: none` and  `tooltip_min` and `tooltip_max` should have the `in` class applied

bafio:
@rovolution sounds reasonable, actually sounds great.
@seiyria @rovolution I'm new to javascript testing, and I'm willing to learn, would you be so kind to point me in the right direction on how to have a proper test environment in order to implement the required test? What I need to install? Which is the lib/dsl used for test?
Thanks a lot!

seiyria:
If you check out the [specs folder](https://github.com/seiyria/bootstrap-slider/tree/master/test/specs), you'll see we use jasmine. You can find an example [here](http://jasmine.github.io/2.0/introduction.html). To test, just do `npm test`.

bafio:
@seiyria thanks a lot

rovolution:
I would add this to the (ConflictingOptionsSpec.js)[https://github.com/seiyria/bootstrap-slider/blob/master/test/specs/ConflictingOptionsSpec.js] file

bafio:
@rovolution will do, thanks for the tip

rovolution:
@bafio after second thought, since this is only really testing `tooltip_split`, i would go ahead and create a new spec file in the specs directory called `TooltipSplitOptionSpec.js`

Then within that file, I would have one big describe block for all the tests, and 2 describe blocks within that...one for when `tooltip_split` is `true` and one for when ``tooltip_split` is `false`

```javascript
// TooltipSplitOptionSpec.js
describe("'tooltip_split' Option tests", function() {
  var testSlider;

  describe("When 'tooltip_split' is true", function() {    
     beforeEach(function() {
         // Do testSlider initialization here
     });

     it("should have `tooltip-main` displayed with `in` class", function() {
     });

     it("should have `tooltip-min, tooltip-max` not displayed", function() {
     });
  });

  describe("When 'tooltip_split' is false", function() {
    beforeEach(function() {
         // Do testSlider initialization here
     });

     it("should have `tooltip-min, tooltip-max` displayed with `in` class", function() {
     });

     it("should have `tooltip-main` not displayed", function() {
     });  
  });
  
  afterEach(function() {
      // Do testSlider teardown here
  });
});
```

Otherwise, the actual tests you wrote look pretty solid! good work! Make those changes and then @ mention me.

bafio:
@rovolution test moved, thanks for the suggestions

rovolution:
looks good! Thanks for the PR @bafio!

bafio:
@rovolution you're welcome, thanks for the great plugin

Issue #369, opened 2015-05-04
Slider won't keep modified value on slideStop


deathwebo:
Hi there!

I found that if I change the value on the fly of my slider on the **"slideStop"** event the value is overwritten right after the event. 

I'm trying to validate that the total amount of some related sliders is not greater than 100.

In case the value of a slider (added to the other slider's values) surpasses the 100 mark, I need to change that slider's value to the next available value. So if a slider is changed to 60, and the total of the rest of the sliders is 50, then I have to change the 60 value to 50.

Here's a working example of my bug:
https://jsfiddle.net/deathwebo/roks4ny0/

I found out on the code, in the **mouseup** and **keydown** events where the slideStop is triggered that if I change this piece of code
```
this._trigger('slideStop', val);
this._setDataVal(val);
```

To this 
```
this._setDataVal(val);
this._trigger('slideStop', val);
```
It gets fixed as the value is not overwritten by the _setDataVal method.
https://jsfiddle.net/deathwebo/Lvbq9uh8/

Cheers! And keep up the good work @seiyria & co

seiyria:
I agree with your assessment; I think it's wise to emit the event after having changed the data (as that's what I would expect in an event emitter situation). If that would suffice, I think that change is simple enough to implement. If you send a PR, that'd be great too!

PR #370, opened 2015-05-05
Changed the order of the slideStop event, moved right after dataSetVal issue #369

deathwebo:
Small change to trigger the slideStop event after the setDataVal, instead of before.

This fixes the issue #369

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1187,8 +1187,8 @@
 				this._setDataVal(val);
 				this.setValue(val, true, true);
 
-				this._trigger('slideStop', val);
 				this._setDataVal(val);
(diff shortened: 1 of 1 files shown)


seiyria:
This seems simple enough. @rovolution - what do you think?

rovolution:
If we could get some sort of automated test that could check to see if the `value` attribute of the `<input>` element associated with the slider got updated within a `slideStop` event listener, that would be great!

You can create a new spec file called `SlideStopEventSpec.js` in our `/specs` directory

rovolution:
After reconsideration, I think ill just merge this in because it is not breaking the existing test suite, and simulating this behavior with JS is somewhat challenging.

PR #380, opened 2015-06-03
CSS Fix: vertical orientation in reverse(RTL) mode

vukasin0:
When slider work in *revered: true* mode and general orientation of page is RTL, ticker text has margin-left css property. To display it properly it should have margin-right.


seiyria:
Your build failed. Can you please fix that, and also show a before and after jsfiddle so we can see what you did?

Additionally, it may be nice to have a unit test for this particular case so going forward this can't be broken again.

Thanks!

rovolution:
going to close for now since the tests are failing and this PR hasnt been active for a while

PR #385, opened 2015-06-11
Make slider ticks 100% opaque.

richiecarmichael:
Previously slider ticks were 80% opaque, on dark backgrounds the color dependency between the opaque track and semi-transparent ticks looks kinda weird.

### DIFF
--- a/css/bootstrap-slider.css
+++ b/css/bootstrap-slider.css
@@ -218,7 +218,7 @@
   -moz-box-sizing: border-box;
   box-sizing: border-box;
   filter: none;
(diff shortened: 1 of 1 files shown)


seiyria:
I think this is more of an application-dependent thing. In general, what we have looks good... I'd leave it up to you to style it as you want.

That's just my opinion though. What do you think @rovolution?

rovolution:
@richiecarmichael I would agree with @seiyria that style is an app-specific thing. we try to encourage people to use CSS to style it however they wish

PR #390, opened 2015-06-22
Add insertBefore bool option

moshfeu:
Now, the user can choose if the slider will create before or after the input and hide or not the original element. It's good if you want to show  the element and show the slider as popup.

Demo: https://jsfiddle.net/4hj4Lq0w/3/

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -448,8 +448,7 @@
 				var sliderTooltipMax = document.createElement("div");
 				sliderTooltipMax.className = "tooltip tooltip-max";
 				createAndAppendTooltipSubElements(sliderTooltipMax);
-
-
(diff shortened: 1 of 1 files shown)


seiyria:
I have literally never seen this use case before, and I'm not sure it's valuable. It seems more confusing than anything.

moshfeu:
That's why I attached a demo =)

seiyria:
Right, I saw the demo and it still took a while. I am personally not seeing the use case still, but that aside; for us to accept this, we would prefer if you added tests, an example, and documentation since this is a brand new feature.

rovolution:
yea i am little confused by this use case (and I do not think it is general enough to adapt to a broader range of cases). Going to close this PR for now

PR #394, opened 2015-06-24
Fixes #392

### LINKED ISSUE #392: Snap doesn't work with ranges
The slider doesn't snap to the closest tick when `range` and `ticks_snap_bounds` are set to `true`.

This is because the value is an array of two elements when the slider is a range. As a consequence, at [line 1286](https://github.com/seiyria/bootstrap-slider/blob/master/js/bootstrap-slider.js#L1286), we actually compare a number with an array :

``` javascript
_calculateValue: function(snapToClosestTick) {
    var val;
    if (this.options.range) {
        // Here we do val = [something, something]
    } else {
        // Here we do val = something
    }

    if (snapToClosestTick) {
        var min = [val, Infinity];
        for (var i = 0; i < this.options.ticks.length; i++) {
            // Number - Array === NaN
            var diff = Math.abs(this.options.ticks[i] - val);
            if (diff <= min[1]) {
                min = [this.options.ticks[i], diff];
            }
        }

    // etc.
}
```

To fix this we need to know which of the two sliders is being moved. I haven't browsed the code enough yet to know if it's possible and how.

kjaquier:
See issue #392.

I also modified the example 14 in the documentation so that it shows this case is possible.

### DIFF
--- a/index.html
+++ b/index.html
@@ -713,7 +713,7 @@ var slider = new Slider("#ex13", {
 
 	<div class='slider-example'>
       <h3>Example 14:</h3>
-      <p>Using tick marks at specific positions..</p>
+      <p>Using tick marks at specific positions, with a range slider.</p>
       <div class="well">
       <input id="ex14" type="text"/>
       </div>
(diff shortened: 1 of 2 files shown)


seiyria:
Seems simple enough. I think a case like this would require a test just to verify that these options are compatible in the future, so it doesn't break again. After that, I think it looks great!

js/bootstrap-slider.js
```diff
@@ -1301,15 +1301,19 @@
 				}
 
 				if (snapToClosestTick) {
-					var min = [val, Infinity];
 					for (var i = 0; i < this.options.ticks.length; i++) {
-						var diff = Math.abs(this.options.ticks[i] - val);
-						if (diff <= min[1]) {
-							min = [this.options.ticks[i], diff];
+						var tick = this.options.ticks[i], draggedVal, snappedVal;
```
rovolution: could you declare each of the `var`s on a different line?

seiyria:
@rovolution After that style change is made, do you think this is good to go?

index.html
```diff
@@ -713,7 +713,7 @@
 
 	<div class='slider-example'>
       <h3>Example 14:</h3>
-      <p>Using tick marks at specific positions..</p>
```
rovolution: these changes need to happen in the [index.tpl](https://github.com/seiyria/bootstrap-slider/blob/master/tpl/index.tpl) file.

that file compiles down to this file

PR #395, opened 2015-07-01
added bootstrap like css classes also a ticks_classes option

Bhoft:
Hi i have added bootstrap like CSS classes slider-danger, slider-warning, slider-success and slider-info in the css (as an example)

And did some modifications to the js code:
The CSS class of the input element will be used for the slider.

Also the CSS class can automatically set by the selected value where the classname could be set by the "ticks_classes" attribute.
e.g.
```
<input type="text" id="expertise" name="expertise" value="2" data-slider-value="2" class="slider-success">
$('#expertise').bootstrapSlider({'ticks':[1,2,3],'ticks_labels':['Weak','Medium','Strong'],'ticks_classes':['slider-danger','slider-warning','slider-success'],'set_tick_classes':'greatest','ticks_snap_bounds':1});

```

The attribute "set_tick_classes" is for either check for the "greatest" or "smallest" tick value to use the css class of the tick for the slider.

slider_examples


seiyria:
You'll definitely want to fix the build at the very least.

seiyria:
Okay, so it looks like you've added new options. This means that we'll definitely need documentation updates, an example (on the gh-pages branch), and preferably some unit tests on this just so we can verify that this feature isn't broken going forward, and that all cases are currently covered as well.

I think this is a a good feature.

css/bootstrap-slider.css
```diff
@@ -243,4 +244,75 @@
   background-repeat: repeat-x;
   filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#ff89cdef', endColorstr='#ff81bfde', GradientType=0);
   opacity: 1;
-}
+  background-color:#89cdef;
```
rovolution: you need to add these to the LESS and SASS files (which is the source that compiles into the css)

js/bootstrap-slider.js
```diff
@@ -1469,6 +1480,59 @@
...
+			},
+			_setTicksCSSClass: function(index) {
+				if(typeof this.options.ticks_classes[index] !== 'undefined') {
+					// remove all the classes
+					this._removeClass(this.sliderElem, this.options.ticks_classes.join(" "));
+					// set new classname
+					this._addClass(this.sliderElem,this.options.ticks_classes[index]);
+				}	
+			},
+			_setTicksClass: function() {
```
rovolution: this is a very large method. why is it needed?

css/bootstrap-slider.css
```diff
@@ -243,4 +244,75 @@
...
+}
+
+.slider.slider-success .in-selection,
+.slider.slider-success .slider-selection.tick-slider-selection,
+.slider.slider-success .slider-handle.min-slider-handle,
+.slider.slider-success .slider-handle.max-slider-handle {
+  background-image: unset;
+  background-color: #62c462;
+  /*background-color: #95d06e;*/
```
rovolution: remove commented code

Bhoft: Those "bootstrap like" css settings where just added as an example.
As you see i am not a huge fan of backgrounds with gradients and therefore i just overruled your settings and unset the background-image. But i think as something like this should be available and at least bootstrap 3 also just uses one simple background-color instead of a gradient those gradient settings and also those bootstrap like css settings should maybe added in an additional files as examples?
This depends on which direction the slider should develop i think and should done by the owner of the extension.

css/bootstrap-slider.css
Bhoft: can't remember that i have added the background-color maybe it was just for some test.

js/bootstrap-slider.js
Bhoft: the _setTicksClass looks which tick is the smallest and which is the greatest tick for the selected values.
and then uses the _setTicksCSSClass to set the new css class which should be set for which tick.

the commented code is also just some "test" to get the closest tick. so that the tick css class will be set if the value is closer to one tick than to another tick. but i am not sure if this is usefull at all therefore it is commented.

the code currently have 2 setting possibilities "greatest" and "smallest" but the "closest" option could also be implemented easily.

PR #397, opened 2015-07-02
Fixes #85

### LINKED ISSUE #85: Bug when both using range slider and reversed
When using range slider with the property "reversed" set to True, the slider is not behaving properly.
I made a quick jsfiddle to show this problem (it functions perfectly with single value slider, but not with range slider): http://jsfiddle.net/XV8B5/

Nu-hin:
Fixed bug with 'range' and 'reversed' options enabled together. 
See an example here: http://jsfiddle.net/p2gvow3k/3/

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -944,7 +944,7 @@
 				var positionPercentages;
 
 				if(this.options.reversed) {
-					positionPercentages = [ 100 - this.percentage[0], this.percentage[1] ];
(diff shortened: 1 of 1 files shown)


seiyria:
Your build failed. Additionally, if possible, it would be great to have a test to cover this case so it doesn't come up again!

Nu-hin:
I updated the fiddle http://jsfiddle.net/p2gvow3k/3/

rovolution:
@Nu-hin could you write an automated test for this?

rovolution:
I went ahead and wrote an automated test for this

PR #402, opened 2015-07-15
Avoid creation of tooltip elements when options.tooltip === 'hide'

Aumnia:
Hi,

Great component, I just wanted a quick way to avoid the creation of the tooltip elements when they were not being used.

Thanks,
/Carl

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -436,28 +436,31 @@
 
 				};
 
-				/* Create tooltip elements */
-				var sliderTooltip = document.createElement("div");
-				sliderTooltip.className = "tooltip tooltip-main";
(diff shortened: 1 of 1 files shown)


seiyria:
Any particular reason for this? Also, if your code doesn't pass the build, we can't merge it!

Aumnia:
No problem I'll try to find some time to determine which tests are assuming tooltip support.

The only reason is to avoid writing markup (DOM elements) that are extraneous when tooltips are not being used.

I realize it was not an elegant addition but maybe in a future refactor you could add this feature.

rovolution:
@Aumnia I think this is a great idea! if you can fix the tests and add automated test coverage for this, i would be all for it

Issue #407, opened 2015-07-28
Refreshing is not working correctly


action-this-day:
When I use $("element").slider("refresh"), the slider refreshes but the value is getting weird.
Instead of the values 1, 2 and 3 I get values like 1,1, 1,2 and 1,3.

Update:
jsfiddle: http://jsfiddle.net/sntcry83/1/

seiyria:
Please make a jsfiddle.

action-this-day:
I've added a jsfiddle

severinmoussel:
Just found another weird behaviour in this fiddel, the refreshed slider became a range slider.

seiyria:
I thought that WAS the weird behavior.

action-this-day:
That's probably why the value is 1,2 instead of just 2 for example.
Why would it change to a range slider?

severinmoussel:
I found that this.options.value is transformed to an array (lines 747 and 749), but I don't see why. And in createNewSlider (line 571), if this.options.value is an array, the slider automatically becomes a range slider.

rovolution:
Thank you for reporting this! Could you list out the steps to re-create each of these bugs? I'd like to start by covering them with automated tests to ensure that they won't reoccur once a fix is made

action-this-day:
What do you mean by "list out the steps to re-create each of these bugs"?
The jsfiddle demonstrates the problem and how to re-create it.

severinmoussel:
Just sent a Pull Request with a possible fix.
The JS Fiddel works well with it.

action-this-day:
I honestly have no idea how to create a pull request. Also, I don't have a fix for this problem, I just know it's there.

severinmoussel:
Sorry, I meant that I just sent one ;)

action-this-day:
Ah haha. My mistake. Thanks!

rovolution:
fixed in v4.14.5

PR #408, opened 2015-07-31
Issue 407 - Refreshing is not working correctly

severinmoussel:
Changed the way this.options.value is handled while not in range mode.
Ensure this.options.scale Object stays as is.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -508,7 +508,7 @@
 			this.tooltip_max = this.sliderElem.querySelector('.tooltip-max');
 			this.tooltipInner_max= this.tooltip_max.querySelector('.tooltip-inner');
 
-			if (typeof this.options.scale != 'Object' && SliderScale[this.options.scale]) {
+			if (typeof this.options.scale !== 'Object' && SliderScale[this.options.scale]) {
 				this.options.scale = SliderScale[this.options.scale];
 			}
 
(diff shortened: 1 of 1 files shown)


seiyria:
I was just about to say to fix your code, but you committed before I could comment! Overall, this seems fine - are you able to throw in a test or two to prevent this behavior from happening in the future as well?

severinmoussel:
Sorry, I don't have much time to do so. And I'm not familiar with JS unit tests.

rovolution:
@seiyria lets wait to get some automated tests before merging

severinmoussel:
I think the best tests would be to do all other tests again after a refresh.

seiyria:
I think it could actually be done more simply:

* create normal slider
* call refresh
* verify it isn't a range slider (and it has the right value)

do the same thing for a range slider.

severinmoussel:
If this parameter has changed between init and refresh, that could happen to any other parameter in the future.

The best practice is to avoid using options for anything else than an option. this.options.value must always represent the "default" value. The current value should be stored elsewhere.

rovolution:
@severinmoussel i definitely agree with you that the current slider value should probably be stored separately than the option. I am attempting to do that refactor at the moment

rovolution:
@severinmoussel I am refactoring how we store internal state within the slider and this should fix your issue.

I will write an automated test to verify this issue is fixed. Ill let you know when im done

rovolution:
https://github.com/seiyria/bootstrap-slider/issues/407

PR #409, opened 2015-08-02
Refactor internal state; Test coverage for refresh() error 407

rovolution:
@seiyria; @severinmoussel this should resolve https://github.com/seiyria/bootstrap-slider/issues/407 and should better organize our internal state.

let me know if it all looks good and ill merged in.

Here is the updated fiddle to verify that #407 is resolved by this: https://jsfiddle.net/sntcry83/6/

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -304,6 +304,22 @@
 
 		function createNewSlider(element, options) {
 
+			/*
+				The internal state object is used to store data about the current 'state' of slider.
+
+				This includes values such as the `value`, `enabled`, etc...
+			*/
+			this._state = {
+				value: null,
+				enabled: null,
+				offset: null,
+				size: null,
+				percentage: null,
+				inDrag: null,
+				over: null
+			};
+
+
 			if(typeof element === "string") {
 				this.element = document.querySelector(element);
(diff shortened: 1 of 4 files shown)


seiyria:
I like the idea of having a state instead of manipulating options repeatedly. This looks good to me.

rovolution:
yea @severinmoussel's comment about options representing options-only made a lot of sense, but i wanted a namespace that would allow us to explicitly indicate internal state variables

severinmoussel:
it's really the cleanest way to do it.

severinmoussel:
But, there is a lottle bug I saw coming. When you refresh the slider, some elements in the state must be kept.

For example, when I call refresh, I think the currently selected value must be kept. Only visual must be refreshed. Otherwise, its not a "refresh" but a "reset"

rovolution:
@severinmoussel I think that would make it really complicated / convoluted. I think the "reset" effect is the most simplest and non-confusing. Thoughts @seiyria?

seiyria:
I'm in favor of having a documented solution for both reset and refresh.
Refresh would do just options and reset would reset the value to the
initial state. Otherwise to set numerous options you'd have to set them one
by one instead of all at once. I can see a use case for wanting each and
having them separate.

Issue #412, opened 2015-08-06
<input> get's attribute data="value: '-5'"


kozlitinaelja:
When run w3c validator this causes error "Attribute data not allowed on element input at this point." If it possible, can you re-name it to data-value="value: '-5'"  and adjust js dependencies to rely on data-value instead of data it would solve the problem.

Thank you, 
Elnara

seiyria:
If I understand you correctly, you can already specify `data-slider-value`, or simply pass it in when you initialize it via JS.

kozlitinaelja:
I mean that data="value: '-5'" which is added via javascript isn't legal attribute on input field based on W3C validator
<img width="1258" alt="screen shot 2015-08-06 at 12 12 42 pm" src="https://cloud.githubusercontent.com/assets/3871309/9116840/8359d172-3c34-11e5-8e4a-1a61617cfaa4.png">

seiyria:
I see. Sorry, to me this is not something that's very important. If @rovolution thinks it's worth pursuing we can reopen this, though. You are, of course, always welcome to submit a PR (but that doesn't come with any guarantees either).

rovolution:
@kozlitinaelja thanks for catching this! If you could submit a PR with an automated test verifying this is correct, that would be much appreciated. thanks!

PR #413, opened 2015-08-08
Convering 'data' attribute to be 'data-value' and directly set equal to the value

rovolution:
Should resolve https://github.com/seiyria/bootstrap-slider/issues/412

What are your thoughts @kozlitinaelja?


rovolution:
alright im merging this in and closing the issue

Issue #232, opened 2014-10-12
Tooltips do not hide on Firefox


lukaszgrolik:
I'm using Firefox 32.0.3 on Mint 17, but this issue probably affects Ubuntu as well. I don't know about Windows.

![seiyria-bootstrap-slider-firefox-issue](https://cloud.githubusercontent.com/assets/3110622/4605542/d3a8685a-51e8-11e4-8610-151eb4a1e38c.gif)

Tooltips disappear only when scrolling the page. 

[JSFiddle](http://jsfiddle.net/as502ak6/1/)

pelegm:
I confirm this bug on Ubuntu 14.04, firefox 37.0.1.

kmkroski:
I'm seeing the same issue on Chrome 44 on OSX 10.10.4.

kmkroski:
I've found a fix: on line 318, set inDrag to false. 
`inDrag: false,`

seiyria:
Want to submit a PR @kmkroski?

AidasK:
Fix is working, merge please :+1:

seiyria:
If we don't have a suitable test to back it up, a fix doesn't really mean
much.

AidasK:
But it is a must, I cant use this library without it. And maybe this fix won't break any of existing tests.

seiyria:
It's not that this fix will break any existing tests, it's that this fix, without a test, can be broken later.

AidasK:
But it could work now and it is broken in current version anyway.

seiyria:
Then please feel free to submit a PR with an accompanying unit test so, going forward, this doesn't happen again.

Bhoft:
He should write a test for some bug which was introduced by @rovolution by this commit https://github.com/seiyria/bootstrap-slider/commit/4942c1bc393d7051d82f482e318fc669e4953536 and wasn't captured by a unit test because he wanted to have one parameter set to false instead of null.
This is strange.... btw please also then fix the _state.over which is also set to null by default. Both parameters had default values of false in the older versions btw.

rovolution:
@Bhoft can you not just create your own fork, make the fix, and then use your fork for the time being? That is the whole point of open source.

Bhoft:
@rovolution of course ;) i never wanted to sound unthankful for your work.
I also already fixed it in my fork ;)

seiyria:
[...]

PR #414, opened 2015-08-08
Fix Tooltip #232

Dennis14e:


seiyria:
Please provide a jsfiddle so we can verify this fix.

Dennis14e:
Without fix: http://jsfiddle.net/Dennis14e/z4saj0L4/
Fix: http://jsfiddle.net/Dennis14e/rj41c4co/

seiyria:
Seems fine to me. @rovolution?

rovolution:
I would like to see an automated test that fails in the broken version/passes in the working version

Dennis14e:
I don't know how to do it.
I've just taken the fix from the Issue - and it works.

rovolution:
> I don't know how to do it.

Please look at our existing test suite as guidance. Automated tests are crucial as they allows us to ensure regressions do not occur in the future

Dennis14e:
I've tried it, but i don't get it - I still don't know how this should work.
You can see the bug in the old version very clearly - and in my version it works.

rovolution:
included as part of @Bhoft's PR: https://github.com/seiyria/bootstrap-slider/commit/90e89c93679affd2d8b7ca2afaa0eb9bd53fead4

PR #416, opened 2015-08-10
Added moveSlider function

mstagg:
Ended up using slider to control playback for an asset tracker. Couldn't find method to update position of slider without direct user input. Wrote one.


seiyria:
Could you please provide a jsfiddle to show us how this works, and add a unit test? Thanks!

rovolution:
yea I would like to see an example of this via a Fiddle.

PR #430, opened 2015-08-22
Update README.md

zedtux:
This MR is about to point the RoR link to the right repository


seiyria:
Sounds good to me, thanks.

PR #432, opened 2015-08-31
Fix finger swipe problem on IE and Windows 8.1 applications

cell42:


seiyria:
What issue(s) does this address and can you make any tests to cover this so we don't break it again later?

seiyria:
Also, does it make sense to to assign the listener twice? Why not look for the correct event and listen to that?

cell42:
I add issue on this address:
https://github.com/seiyria/bootstrap-slider/issues/433
I tested this on Windows Surface device.
This is good idea to recognize type of browser and then attach proper event. But some of people can use touch and mouse at the same time and if You add only one event type then for example mouse swipe may not work.

dist/css/bootstrap-slider.css
```diff
@@ -30,6 +30,7 @@
   display: inline-block;
   vertical-align: middle;
   position: relative;
+  -ms-touch-action: none;
```
rovolution: could you add this to the https://github.com/seiyria/bootstrap-slider/blob/master/less/rules.less file instead?

This is a compiled file

rovolution:
see my feedback on the code.

could you add some automated tests for this?

PR #437, opened 2015-09-04
fixing tooltips on mouse leave

Bhoft:
https://github.com/seiyria/bootstrap-slider/issues/232

hopefully everything is correctly done. never created such tests before, but test fails if null is set.


seiyria:
@rovolution - tests pass, they look ok. This seems ok to me.

rovolution:
@Bhoft after that typo, ill merge and bump the version

Thank you for making the PR and adding the appropriate test coverage!

PR #441, opened 2015-09-10
Add style field to package.json

ehuisman:
In order to use bootstrap-slider with tools like [Parcelify](https://github.com/rotundasoftware/parcelify) and [Sheetify](https://github.com/sheetify/sheetify), the `package.json` file needs a `style` field referring to the correct CSS file. Bootstrap itself, for example, already has this.


seiyria:
Thanks! Seems pretty simple.

PR #444, opened 2015-09-11
only remove tooltip event handlers when they exists

HeikoH:
(removing an `undefined` listener causes angular2 to blow up)


seiyria:
Please fix your code to pass the build and also add unit tests to cover this case so it doesn't happen again. Thanks!

rovolution:
@HeikoH I went ahead and refactored this a little bit/added in more test coverage

PR #453, opened 2015-09-28
- fixes #439 : ticks are not reversed when reversed=true

### LINKED ISSUE #439: Tick start from bottom in vertical slider
Hi,

I am using vertical slider. By default Tick starts from the top. Is there any way i can change it to start from the bottom. Something like desired section in the image.

image

Thanks.

gormanb:
Simple fix to reverse the tick positions when "reversed=true" has been set in the slider options.


seiyria:
Please add a unit test so we can have this not be a problem going forward, and a jsfiddle so we can check it out. Thanks!

rovolution:
I fixed a couple things with this + added automated tests: https://github.com/seiyria/bootstrap-slider/pull/459

PR #456, opened 2015-09-30
Fix tick mark position in logarithmic scale

### LINKED ISSUE #419: Tick marks not aligning on logarithmic scale
The tick marks and labels are not matching the value on the bar when using logarithmic scale. The ticks at least seem to be using linear scale still while the labels are just evenly spaced across the bottom, not matching value.

http://jsfiddle.net/ppdtt6cz/1/

As you can see the tick mark for 50 says it is value 20.

kellenp:
Fixes #419 

Hi there - I changed the tick positions to follow whatever percentages the current scale uses - it was previously only positioning ticks by linear scale.

Here's a JSFiddle with the fix shown
https://jsfiddle.net/ppdtt6cz/3/

I added a test, and also made a small edit to another logarithmic test description to better indicate what it is testing.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -980,8 +980,6 @@
 
 				/* Position ticks and labels */
 				if (Array.isArray(this.options.ticks) && this.options.ticks.length > 0) {
-					var maxTickValue = Math.max.apply(Math, this.options.ticks);
-					var minTickValue = Math.min.apply(Math, this.options.ticks);
 
 					var styleSize = this.options.orientation === 'vertical' ? 'height' : 'width';
 					var styleMargin = this.options.orientation === 'vertical' ? 'marginTop' : 'marginLeft';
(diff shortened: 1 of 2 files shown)


seiyria:
This looks like a good PR. Thank you!

@rovolution - what do you think? Mergable?

kellenp:
JSFiddle may be caching somehow but its possible the initial load may not be effected by this fix - looking into.

kellenp:
Yup, JSFiddle was caching some test data I had filled in, looks fine as best as I can tell.

rovolution:
@seiyria this has been a busy week for me but ill try to review all of these PRs sometime in the next few days

seiyria:
No worries. I try to filter them before they get to you at least.

rovolution:
thanks man! ill take a look this Sat

rovolution:
alright just pulled down @kellenp's branch and ran through the examples again...everything looks good. and thank you for adding in an automated test @kellenp!

PR #463, opened 2015-10-12
Fixed max value selection

arminkarimi:
When we have a slider with a custom step value, if the max slider value is not equal to a multiple of the step value, slider fails to select the max value.
For example, having the following settings prevents the user to select the maximum value:
{
min:0,
max: 4200,
value:1000,
step:1000
}

Here is the example : http://embed.plnkr.co/D3GJQn/preview

If bootstrap-slider-fixed.js is referenced, it will work.


seiyria:
Please add unit tests so we can prevent this from happening in the future, and also rebase your commits when you're done, as you appear to have to commits with the same content.

arminkarimi:
Hi,

I tried my best to implement your comments. Can you please point me to the right direction if what I have added is not correct?

Thanks,

seiyria:
I think the unit test isn't right, but I'm not sure. @rovolution?

arminkarimi:
Sorry. I am not familiar with JS testing frameworks. It would be helpful if there is a read me file (maybe in CONTRIBUTING.md) about the testing library/framework, so people like myself can implement proper tests.

seiyria:
Sure, no problem. I updated the CONTRIBUTING file. For future reference, we use Jasmine.

rovolution:
@seiyria i would mention in the contributing doc that we use the 1.3 version of jasmine (and link to the docs): http://jasmine.github.io/1.3/introduction.html

seiyria:
@rovolution thanks! I wasn't sure what version it was and it was hard to deduce.

test/specs/ValueSelectionSpec
```diff
@@ -0,0 +1,42 @@
+/*
```
rovolution: the file needs the `.js` extension

arminkarimi:
I added the extension but now there is an error in the test file I guess. I am still trying to wrap my head around Jasmine.

arminkarimi:
The test is now passed but I am not sure if the test is actually capturing the mouse drag event. @rovolution any direction that I can make sure test is capturing the scenario?

js/bootstrap-slider.js
```diff
@@ -1371,11 +1371,16 @@
...
-				var percentage = (distanceToSlide / this._state.size) * 100;
-				percentage = Math.round(percentage / this._state.percentage[2]) * this._state.percentage[2];
-				if (this.options.reversed) {
-					percentage = 100 - percentage;
-				}
+				var selectedPercentage = (distanceToSlide / this._state.size) * 100;
+		                var nearestPercentage = Math.round(selectedPercentage / this._state.percentage[2]) * this._state.percentage[2];
+		
+		                // if the maximum is not equal to a step value, select max when user selects the max value.
+		                var percentage = 100 - selectedPercentage < Math.abs(nearestPercentage - selectedPercentage) ?
```
rovolution: could you refactor the boolean `100 - selectedPercentage < Math.abs(nearestPercentage - selectedPercentage)` into a variable describing what the condition is, and then reference it here?

test/specs/ValueSelectionSpec.js
```diff
@@ -0,0 +1,42 @@
...
+  *************************
+
+  This spec has tests for checking if a specific value can be selected, given certain slider options.
+  
+*/
+describe("Value Selection Tests", function() {
+
+  var testSlider;
+
+  it("Should slide to the 4200 when slider handle is dragged to the right most.", function() {
```
rovolution: so this test is not actually doing what this description says.

Currently, the code in this test is doing the following:
- Initializing a slider that starts at the value `4200`
- Creating a mouse event with no positional data specified
- triggering the mouse event
- The test is passing because you initialized the slider at the maximum value. so `expect(slideEvt.value).toBe(4200)` will always pass.


What you want it do is this:
- Initialize the slider at some starting `value` that is not the maximum
- Create a mouse event with positional information specified that would slide the handle to the rightmost point
- then call the `getValue` method of the slider, and assert to see if the slider is at the maximum possible value

PR #470, opened 2015-10-30
Fix label positions when using slider in vertical orientation.

sesh-kebab:
Using the slider in vertical orientation with labels has the labels rendering in the incorrect location. The fix renders the labels in a more appropriate position when using the slider in vertical orientation.

image

Original
http://jsfiddle.net/seshi/mh3d14ze/

Fix
http://jsfiddle.net/seshi/smogd9mk/

### DIFF
--- a/css/bootstrap-slider.css
+++ b/css/bootstrap-slider.css
@@ -104,6 +104,12 @@
   border-left-color: #0480be;
   margin-left: 0;
 }
+.slider.slider-vertical .slider-tick-label-container {
+  white-space: nowrap;
+}
+.slider.slider-vertical .slider-tick-label-container .slider-tick-label {
+  padding-left: 4px;
+}
 .slider.slider-disabled .slider-handle {
   background-image: -webkit-linear-gradient(top, #dfdfdf 0%, #bebebe 100%);
   background-image: -o-linear-gradient(top, #dfdfdf 0%, #bebebe 100%);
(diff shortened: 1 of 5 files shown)


seiyria:
Please provide an adequate test to support this not happening again and a jsfiddle so we can take a look at it.

sesh-kebab:
@seiyria thanks! I just found contributing.md; so feeling a bit silly having submitted a pr without all the required items.

jsfiddle shouldn't be problem but not sure how I can write an adequate test to validate the fix as it's a UI issue. Would you please be able to point me to an example?

seiyria:
[This test file](https://github.com/seiyria/bootstrap-slider/blob/master/test/specs/LowAndHighTrackSpec.js) might be a good one to check out, as it also does UI checking.

sesh-kebab:
Added http://jsfiddle.net/seshi/mh3d14ze/

seiyria:
I think this makes sense and looks ok, but I will defer to @rovolution ultimately on the quality. Regardless, you've made the process much easier, so thank you very much!

rovolution:
Excellent PR @sesh-kebab! Thank you for adhering to the contributing guidelines and making an awesome PR!

sesh-kebab:
@rovolution thanks - glad to be contributing :smile: have another fix for an issue where the labels aren't rendered correctly inside of a popup but will start a separate PR for that (needed this fixed first before I can show the popup issue).

PR #478, opened 2015-11-30
Better accessibility

cornedor:
Currently the values are not read back to screenreaders, I've tried to add some accessibility based off http://www.oaa-accessibility.org/examplep/slider1/.

Example: http://jsfiddle.net/f2xsjc0j/1/

Before:
old_situation

After:
new_situation

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | ticks_snap_bounds | float | 0 | Used to define the snap bounds of a tick. Snaps to the tick if value is within these bounds. |
 | scale | string | 'linear' | Set to 'logarithmic' to use a logarithmic scale. |
 | focus | bool | false | Focus the appropriate slider handle after a value change. |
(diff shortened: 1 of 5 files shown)


seiyria:
This looks very well-done! Exactly what we look for in a PR. @rovolution, is this good to merge?

rovolution:
Echoing @seiyria's sentiment...this is a first-class pull request. Excellent work @cornedor!

cornedor:
Nice to hear that and thanks for the quick merge.
I see that you've updated the index.html while there is a line in .gitignore for index.html, is that correct?

seiyria:
That is correct. `.gitignore` only prevents files from being added, not explicitly committed to or force-added.

rovolution:
yea we mostly did that to force people to update the `.tpl` file. I will add compiling and committing the `index.html` as part of the postpublish build script (when I find time here, I am going to move everything from that script to Travis so that is all done as part of the CI process)

cornedor:
Ah I see :-)

PR #489, opened 2015-12-09
Update how slider value is adjusted based on ticks positions

luislee818:
When _ticks_positions_ are specified, the _raw_value_ calculated based on the percentage should not be adjusted with the _min_ value.

Changes:
* Do not adjust with base (min) after calculated with ticks positions
* Add example 13.a in index.html for the case above

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -207,25 +207,26 @@
 			linear: {
 				toValue: function(percentage) {
 					var rawValue = percentage/100 * (this.options.max - this.options.min);
+					var shouldAdjustWithBase = true;
 					if (this.options.ticks_positions.length > 0) {
 						var minv, maxv, minp, maxp = 0;
-						for (var i = 0; i < this.options.ticks_positions.length; i++) {
+						for (var i = 1; i < this.options.ticks_positions.length; i++) {
 							if (percentage <= this.options.ticks_positions[i]) {
-								minv = (i > 0) ? this.options.ticks[i-1] : 0;
-								minp = (i > 0) ? this.options.ticks_positions[i-1] : 0;
+								minv = this.options.ticks[i-1];
+								minp = this.options.ticks_positions[i-1];
 								maxv = this.options.ticks[i];
 								maxp = this.options.ticks_positions[i];
 
 								break;
 							}
 						}
-						if (i > 0) {
-							var partialPercentage = (percentage - minp) / (maxp - minp);
-							rawValue = minv + partialPercentage * (maxv - minv);
-						}
+						var partialPercentage = (percentage - minp) / (maxp - minp);
+						rawValue = minv + partialPercentage * (maxv - minv);
+						shouldAdjustWithBase = false;
 					}
 
-					var value = this.options.min + Math.round(rawValue / this.options.step) * this.options.step;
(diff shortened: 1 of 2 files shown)


seiyria:
Can you please provide a jsfiddle for the before and after (preferably with an explanation so this case is understood) as well as a unit test to cover this situation? Thanks!

index.html
```diff
@@ -711,6 +711,43 @@
       </code></pre>
   </div>
 
+  <div class='slider-example'>
```
rovolution: these changes need to be added to the `index.tpl` file (which compile down to this file)

js/bootstrap-slider.js
```diff
@@ -206,7 +206,9 @@
 		var SliderScale = {
 			linear: {
 				toValue: function(percentage) {
-					var rawValue = percentage/100 * (this.options.max - this.options.min);
+					var rawValue = percentage/100 * (this.options.max - this.options.min),
+						shouldAdjustWithBase = true;
```
rovolution: please declare this var on its own line

rovolution:
> Can you please provide a jsfiddle for the before and after (preferably with an explanation so this case is understood) as well as a unit test to cover this situation? Thanks!

Echoing @seiyria's sentiments...I'd like to see automated test coverage

luislee818:
Thanks for the feedback.

About the tests, can I expose some objects in the closure for testing purpose? In this case, I would like to test the _toValue_ function on the _liner_ property of _SliderScale_ variable, as that's where the change I would like to make. Could _SliderScale_ be attached to _Slider_ variable, like this?
```javascript
Slider.Scale = SliderScale;
```

Or every test has to be done from the outside, on the DOM?

rovolution:
>> Or every test has to be done from the outside, on the DOM?

I would prefer DOM level assertions, simply because that is a better/more reliable test of the behavior

luislee818:
Below are some examples showing the issue and the fix:

Version | Scenario (Tick starting from) | Result (dragging & clicking) | Demo
------------ | -------------| ------------- | -------------
v5.3.1 | 0 | working | http://codepen.io/luislee818/pen/NxqOWr
v5.3.1 | positive value | not working | http://codepen.io/luislee818/pen/XXmMpj
v5.3.1 | negative value | not working | http://codepen.io/luislee818/pen/vLNxgj
pull request | 0 | working | http://codepen.io/luislee818/pen/OMypWr
pull request | positive value | working | http://codepen.io/luislee818/pen/gPammQ
pull request | negative value | working | http://codepen.io/luislee818/pen/NxGppZ

I've added tests covering those three cases and updated implementation.

test/specs/TickClickingBehaviorSpec.js
```diff
@@ -0,0 +1,101 @@
+describe("TickClickingBehavior", function() {
+	var SLIDER_ID = "testSlider1";
```
rovolution: this spec file is somewhat difficult to read.

Could you break each of these scenarios into their own explicit `it()` statement? I know it might be more repetition but it makes the file much more readable/easier to understand

luislee818:
@rovolution I updated the spec.

rovolution:
@luislee818 thanks! Merging/bumping version

rovolution:
done! thanks again for the contribution @luislee818!

PR #490, opened 2015-12-16
Adding new LESS variables and some general cleanup

slindstr:
Hi!

First off thanks a bunch for writing this slider - it's really awesome! 

I need to use this in a project but I need to deviate from some of the colors you have specified as defaults. I figured that rather than overwriting CSS declaration blocks in my project or changing the values in the dist CSS file I'd add some LESS variables. (I also saw [Pull Request 393](https://github.com/seiyria/bootstrap-slider/pull/395) but since it hasn't been touched since July I figured master has probably diverged quite a bit...)

I also did some general cleanup including:

* changing tabs to spaces
* adding spaces around division signs
* making sure each selector is on its own line
* and adding spaces between each declaration block 

The above changes seem more readable to me, but in the event you don't like them lemme know and I'm happy to revert any or all of them. Each of those changes has been made as a separate commit so it's easier for you to review the code.

I have to run home from work at the moment so I haven't had *too* much time to test things, but since everything is a pretty straightforward 1 for 1 switch everything should be good. I'll have more time to look at this stuff tomorrow though.

Thanks again and let me know what you think!


seiyria:
I don't think we really need a general style change in our codebase - if so, that should be:

* submitted as a separate PR
* rebased into one commit

As for the actual content here, I don't see the point - if you're just changing colors, that's an override you should be doing client side. Unless it makes a drastic difference on our end (which, based on the hex colors, it seems to be a very minor color change), it doesn't make sense to PR this.

The only thing that I think makes sense here is abstracting some bits to variables, but that only makes sense if they're used in multiple places, which they aren't always. Even then, this needs to be done in both SASS and LESS to preserve parity between the two. This PR, as is, I will have to close as it covers too much ground. If you would like, I would accept an abstraction on the variables as a PR. I would review a code style changing PR but I don't think it's a necessary change honestly so it would probably be rejected.

Issue #464, opened 2015-10-13
Resize issues, not responsive labels!


adameska:
I fixed a bug, your labels aren't responsive at all, can you please incorporate these changes?
line 649:
this.resize = this._resize.bind(this);
line 657:
window.addEventListener("resize", this.resize, false);

line 1134:
_resize: function (ev) {
				this._state.offset = this._offset(this.sliderElem);
				this._state.size = this.sliderElem[this.sizePos];

				this._layout();
			},

seiyria:
If you could send us a PR and a jsfiddle showing that these changes do indeed fix the issue, plus a test or two to verify that it won't happen again going forward, that'd be awesome!

adameska:
I'm having issues importing the bootstrap-slider.js file to the jsfiddle but here's basically what i'm doing:
http://jsfiddle.net/4e07rLz1/7/

seiyria:
Try using this: https://rawgit.com/

adameska:
Updated, forgot to remove the period in the url. You can see when you resize the screen the labels are stuck, but when you click on it then it redraws the labels correctly.

adameska:
Also, is there anyway to default to null? I can't figure out how to initialize without a value selected in case they don't want to rate the app.

seiyria:
There is no way to do that, no. It does not make sense for a slider to not have a value, as it would be confusing to interact with. Most users would wonder what it is.

rovolution:
Fixed in [v5.3.4 and v5.3.5](https://github.com/seiyria/bootstrap-slider/blob/master/CHANGELOG.md#534--2015-12-27)

PR #496, opened 2015-12-24
Bind _layout() function to window.resize event

zsiswick:
A number of solutions have been proposed to make the slider labels responsive, this was the most elegant one I've seen, originally proposed in [this thread](https://github.com/seiyria/bootstrap-slider/issues/464).

A visual example of the change can be seen [here](http://jsfiddle.net/zsiswick/76sywz9d/2/). Please try expanding and contracting the result window. Note the responsive labels work for vertical slider orientation as well.

### DIFF
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -677,6 +677,7 @@
 			}
 			this.sliderElem.addEventListener("mousedown", this.mousedown, false);
 
+			// Bind window handlers
 			this.resize = this._resize.bind(this);
 			window.addEventListener("resize", this.resize, false);
 
@@ -1172,6 +1173,7 @@
 				this._state.offset = this._offset(this.sliderElem);
 				this._state.size = this.sliderElem[this.sizePos];
         this._layout();
+				this._trigger('resize');
       },
 			_removeProperty: function(element, prop) {
 				if (element.style.removeProperty) {
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -1176,7 +1176,7 @@
 				this._state.offset = this._offset(this.sliderElem);
 				this._state.size = this.sliderElem[this.sizePos];
 				this._layout();
-      },
+			},
 			_removeProperty: function(element, prop) {
 				if (element.style.removeProperty) {
 				    element.style.removeProperty(prop);
--- a/js/bootstrap-slider.js
+++ b/js/bootstrap-slider.js
@@ -677,6 +677,9 @@
 			}
 			this.sliderElem.addEventListener("mousedown", this.mousedown, false);
 
+			this.resize = this._resize.bind(this);
+			window.addEventListener("resize", this.resize, false);
+
 
 			// Bind tooltip-related handlers
 			if(this.options.tooltip === 'hide') {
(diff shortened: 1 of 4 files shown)


seiyria:
This looks okay. I'm not sure if it's possible to unit tests with events like this, but it'd be nice to have if it were.

zsiswick:
I'm updating my fiddle, looks like the changes I made didn't stick. Will update you when it's ready to be reviewed again.

zsiswick:
Okay, the fiddle should be working now :)

zsiswick:
Yes, I'm not having much luck with a unit test, more specifically, `$(window).trigger('resize')` has no apparent effect in a headless browser test. Do you have any suggestions on how to access the slider's private functions in Jasmine? I'm not seeing many examples in the existing tests.

js/bootstrap-slider.js
```diff
@@ -677,6 +677,10 @@
 			}
 			this.sliderElem.addEventListener("mousedown", this.mousedown, false);
 
+			// Bind window handlers
+			this.resize = this._resize.bind(this);
+			window.addEventListener("resize", this.resize, false);
```
rovolution: need to remove this event listener in the `destroy()` method

js/bootstrap-slider.js
```diff
@@ -1164,6 +1168,13 @@
 			        }
 				}
 			},
+			_resize: function (ev) {
+				/*jshint unused:false*/
+				this._state.offset = this._offset(this.sliderElem);
+				this._state.size = this.sliderElem[this.sizePos];
+        this._layout();
```
rovolution: fix the formatting

test/specs/EventsSpec.js
```diff
@@ -224,6 +224,16 @@ describe("Event Tests", function() {
 
     });
 
+    describe("Window Events", function() {
+      it("'resize' event is triggered properly and can be binded to", function() {
+        testSlider.on('resize', function() {
+          flag = true;
+        });
+        testSlider.data('slider')._resize();
```
rovolution: since the `resize()` method is not being publicly exposed, I would rather the test not invoke this method.

I saw your comment stating that the you cannot trigger the `resize` event on the global `window` object in the headless browser. Lets just go ahead and remove this test then

rovolution: Actually, try using `window.dispatchEvent( new Event('resize') )` in the test and see if that works. 

Since you are binding to the browser native event with `addEventListener()`, I am guessing that using the jQuery `trigger()` method only triggers jQuery-specific events.

rovolution: Also,  since this fix is intended to fix the positioning of the ticks, could you add some assertions to the test to ensure the ticks are positioned correctly after the `resize` event is triggered on the `window`?

rovolution:
The fix looks solid @zsiswick. I left some feedback in the code

js/bootstrap-slider.js
zsiswick: Done, I manually tested this.

zsiswick:
I tried triggering the window resize via `window.dispatchEvent( new Event('resize') )`, but to no avail. I think the proper way to test this is captured in the new spec I added, basically manually changing the slider's width, then calling the `_resize()` private method, and asserting the dom elements updated styling.

rovolution:
> I tried triggering the window resize via window.dispatchEvent( new Event('resize') ), but to no avail. I think the proper way to test this is captured in the new spec I added, basically manually changing the slider's width, then calling the _resize() private method, and asserting the dom elements updated styling.

Yup no worries! And the tick position assertions in the tests look great! Thanks for doing that!

Merging in

PR #504, opened 2016-01-12
Update bower.json

AlexeyBay:
Correct path to main js file
(only package.json was fixed before)


seiyria:
Thanks for this.

AlexeyBay:
Thanks!
Could you also update a tag, so bower will see updated version and ignore cache, please?

seiyria:
Should be good to go now.

PR #511, opened 2016-01-25
fix event/input value attribute "race condition"

glaszig:
this simple patch sets the input's value attribute
to the new value _before_ triggering events.
it works around the issue of reading the input's 
value attribute (`$(input).val()`) inside an event listener
which would always return the value of the preceding slide step.

so, the following inside an event listener will evaluate to `true`:

```js
event.value.newValue == input.value;
```

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -814,6 +814,7 @@
 				this._layout();
 				var newValue = this.options.range ? this._state.value : this._state.value[0];
 
+				this._setDataVal(newValue);
 				if(triggerSlideEvent === true) {
 					this._trigger('slide', newValue);
 				}
@@ -823,7 +824,6 @@
 						newValue: newValue
 					});
 				}
-				this._setDataVal(newValue);
 
 				return this;
 			},
--- a/test/specs/EventsSpec.js
+++ b/test/specs/EventsSpec.js
@@ -47,6 +47,26 @@ describe("Event Tests", function() {
         expect(flag).toBeTruthy();
       });
 
+      it("'slide' event value and input value properties are synchronous", function() {
+        testSlider.on('slide', function(e) {
+          flag = true;
+          expect(e.value.toString()).toEqual(this.value);
+        });
+        testSlider.slider("setValue", 3, true, false);
+
+        expect(flag).toBeTruthy();
+      });
+
+      it("'change' event value and input value properties are synchronous", function() {
+        testSlider.on('change', function(e) {
+          flag = true;
+          expect(e.value.newValue.toString()).toEqual(testSlider.val());
+        });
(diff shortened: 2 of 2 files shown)


seiyria:
Not sure I see the purpose here. If you're using jquery, you can listen for the `change` event (emitted right below) and use that value. I don't know what use case it fulfills to read the value your way when we already provide it to you.

glaszig:
> Not sure I see the purpose here

avoiding confusion. we were migrating from html5 sliders to this library. our html5 code simply queried the input's value. this patch would have made it possible for us to migrate only by adding the library and changing the event name in our code -- so, more convenience, less confusion.

seiyria:
Sorry, if that's all it is, I don't think I agree. If you already have to change usage of everything, you may as well change the way you get the value as well. I'll let @rovolution have the final call on this, though.

glaszig:
it's more technical than "convenience". the value change should, in this context, be synchronous. if i get the value passed via the event it should be the same as the actual value on the input element itself. it should work like [the standard defines it](http://www.w3.org/TR/html5/forms.html#common-input-element-events).

seiyria:
Alright, in that case, we'll need a unit test to accompany this before we can accept it. I think that makes sense if it's part of the spec.

glaszig:
i'll see what i can do.

rovolution:
I see what the issue is @glaszig, and it makes sense. Thank you identifying this issue

Like @seiyria said, could you provide some sort of unit test? I think a really good one would be:

* Create a slider instance and bind an event listener to the `change` event
* call `setValue()` and have it trigger the `change` event
* In your event listener, check the value of the slider and see if it matches the value of the `newValue` emitted from the `change` event

glaszig:
sorry to bother you but there's something going on which does not appear obvious to me.
the following test is green:

`EventsSpec.js`

```js
      it("'slide' event value and input value properties are synchronous", function() {
        testSlider.on('slide', function(e) {
          flag = true
          expect(e.value.toString()).toEqual(this.value);
        });
        testSlider.data('slider')._mousemove(mouse);

        expect(flag).toBeTruthy();
      });
```

this one fails:

```js

      it("'slide' event value and input value properties are synchronous", function() {
        testSlider.on('slide', function(e) {
          flag = true
          expect(e.value.toString()).toEqual(this.value);
        });
        testSlider.slider("setValue", 3, true, false);

        expect(flag).toBeTruthy();
      });
```
results:

```
Event Tests:: JQuery version:: Mouse Events:: 'slide' event value and input value properties are synchronous: failed
  Expected '3' to equal '1'. (1)
```

glaszig:
nevermind. i figured it out. was just not competent enough for the tooling.
i added a few tests. let me know what you think.

rovolution:
This PR looks great @glaszig! Awesome work!

PR #514, opened 2016-01-27
fix for vertical slider in scrollable elements.

serbiant:
I experienced the issue when my vertical slider was placed into Y-scrollable

element. The slider was at the right place, but mouse position after scrolling was incorrect (lower than real cursor for same pixels as scrollTop element property). So I fixed this. Please rewiew.


seiyria:
Please provide a jsfiddle (before and after) so we can see the error, and also add unit tests.

rovolution:
remove the `dist` files from this PR

serbiant:
OK, I pushed just the change. Will do rebase.
[Fiddle before](https://jsfiddle.net/serbiantx/xxbjmLks/1/)
[Fiddle after](https://jsfiddle.net/serbiantx/r8Lubn2r/3/)

I'm not sure whether we need the unit test for it. But if you think so, I'll add one.
Notice that this fix will work only if scrollable element is positioned. (explained by **offsetParent()** docs). Actually I can check there for every parent's offset (not for only positioned parents), and it will work without positioning. Please share what do you think

seiyria:
Please do add tests, they help prevent regressions.

rovolution:
@serbiant code looks fine. But as @seiyria said, if you could add a simple unit test it would be much appreciated! We really want to prevent any sort of regressions from occurring, and the unit test suite is our best way to ensure that

serbiant:
@rovolution sure I'll do this. Sorry was busy that days.

serbiant:
Check it out! hope you'll find the test appropriate. (I'm not such a jasmine guru :) ). Sorry I can't include revert into rebase so there're three commits.

rovolution:
@serbiant the tests look excellent...some of the best I have seen from a contributor! Thank you for taking the time to craft high-quality unit tests!

rovolution:
one more thing...could you rename the spec from `ScrollableSpec.js` -> `ScrollableContainerSpec.js`, just to be more specific? Thanks!

serbiant:
@rovolution made rename. Always welcome :)

rovolution:
@serbiant so after merging your branch I realized that it broke vertical slider for when they are not inside a scrollable container.

See Example 4.

I am going to revert this PR for now but could you look into that case as well? Thanks!

serbiant:
@rovolution Sorry missed it too. Just [fixed](https://github.com/serbiant/bootstrap-slider/commit/859184dafa9e037318a923e750677b770a5fef85). should I PR again?

seiyria:
Since this one is considered merged, I'm afraid you will have to open a new one with the same contents.

rovolution:
in your new PR, could you also add tests for when the slider is in the non-scrollable container? Thanks!

rovolution:
@serbiant would you mind opening a new PR with the changes? Thanks!

serbiant:
@rovolution the issue that 'broke' default vertical slider - I just haven't considered **body** possible scroll offset. that 'broken' version would work if page isn't scrolled down. I don't think that tests for this case are necessary, but if you really want I can add one.

rovolution:
@serbiant if you could that would be great! Thanks!

PR #516, opened 2016-01-31
Update slider.scss

samcome:
Fix baseBorderRadius
No mixin but just helper variable ?
Work for me now.


seiyria:
Please:

* don't change a bunch of indentation, that makes it hard to read what you actually did
* also modify the LESS file, because we also need to support that for now
* show us a before and after jsfiddle, because this means nothing to me as it is

rovolution:
[Our SASS file was recently refactored in v7.0.0](https://github.com/seiyria/bootstrap-slider/pull/553/files), so I am going to close this PR

PR #519, opened 2016-02-02
_setText: (minor bugfix) removed use of innerText. resolves issue sei…

gio-js:
…yria/bootstrap-slider#518

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1502,7 +1502,9 @@
 			_setText: function(element, text) {
 				if(typeof element.textContent !== "undefined") {
 			 		element.textContent = text;
-			 	}
+			 	} else if(typeof element.innerText !== "undefined") {
+-			 		element.innerText = text;
+-			 	}
 			},
 			_removeClass: function(element, classString) {
 				var classes = classString.split(" ");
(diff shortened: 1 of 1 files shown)


seiyria:
This looks fine and I'm not sure if there's a test that could make sense here. @rovolution?

src/js/bootstrap-slider.js
```diff
@@ -1500,9 +1500,7 @@
 				this.$sliderElem.off();
 			},
 			_setText: function(element, text) {
-				if(typeof element.innerText !== "undefined") {
-			 		element.innerText = text;
-			 	} else if(typeof element.textContent !== "undefined") {
+				if(typeof element.textContent !== "undefined") {
```
rovolution: instead of removing the `innerText` check, could you just swap textContent and innerText in the original code?

That way, textContent will be the default API for setting the text...which is probably what we should be doing [based on this article](http://www.kellegous.com/j/2013/02/27/innertext-vs-textcontent/)

rovolution: @gio-js if you could make this change and then tag me, it should be good to go.

rovolution:
this looks good to me. Thanks for the pr @gio-js!

PR #521, opened 2016-02-04
Fix for vertical slider in Y-scrollable elements after scroll.

serbiant:
Works fine with body scrolled down.
Per your request, added simple test even for the case when <body> element is scrolled down before interacting with slider.


rovolution:
This looks awesome @serbiant, thank you again for this high quality PR!

PR #531, opened 2016-02-24
Auto register data provide

MaZderMind:
This PR activates automatic registering of `<input />` Elements when they are equipped with a `data-provide="slider"` Attribute. This conforms to the way other Bootstrap-Plugins like [bootstrap-datepicker](https://bootstrap-datepicker.readthedocs.org/en/latest/#data-api) work and mimics the Behavior of Bootstrap's own JavaScript-Plugins, ie [Buttons](http://getbootstrap.com/javascript/#buttons) which are enabled by `data-toggle="button"` or similar Attributes.

All Slider-Options can be set with `data-slider-` Attributes, so in most Situations no custom JavaScript-Code would be required anymore.

### DIFF
--- a/test/specs/AutoRegisterDataProvideSpec.js
+++ b/test/specs/AutoRegisterDataProvideSpec.js
@@ -1,0 +1,21 @@
+describe("Auto register data-provide Tests", function() {
+  it("checks that the autoregister Slider was automatically registerd", function() {
+    var $el = $("#autoregisterSlider");
+
+    var sliderInstancesExists = $el.siblings().is(".slider");
+    expect(sliderInstancesExists).toBeTruthy();
+
+    var sliderInstancesCount = $el.siblings(".slider").length;
+    expect(sliderInstancesCount).toEqual(1);
+  });
+
+  it("checks that the autoregistered Slider can be accessed", function() {
+    var $el = $("#autoregisterSlider");
+    
+    expect($el.slider('getValue')).toBe(1);
+
+    $el.slider('setValue', 2);
+
+    expect($el.slider('getValue')).toBe(2);
+  });
+});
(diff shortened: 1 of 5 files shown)


seiyria:
We would definitely need some tests to go with this before we could accept it. I like the idea, though.

MaZderMind:
@seiyria I have no experience in doing Browser-Side Tests. Would you give me some examples of the Kind of things you'd like to see tested? I'll try to add them then.

seiyria:
Basically, for sure, we need a test to prove that this behavior works. I'm not sure if there's anything else to be tested.

seiyria:
@rovolution what do you think?

test/specs/AutoRegisterDataProvideSpec.js
```diff
@@ -0,0 +1,21 @@
+describe("Auto register data-provide Tests", function() {
+  it("checks that the autoregister Slider was automatically registerd", function() {
```
rovolution: could you add `after()` block that calls the destroy() method on this slider just to ensure proper cleanup & no conflicts with other tests?

Other than that, this looks great! Thanks!

rovolution:
PR looks great @MaZderMind! just left some minor feedback on the tests

test/specs/AutoRegisterDataProvideSpec.js
MaZderMind: I actually left that out intentionally, because the Slider is activated on page-load, not inside the test.
So if I destroy it inside the test, the result of other tests might change with the order of the tests, because destroying an auto-slider would change global state.

rovolution: solid justification. lets keep it the way it is then

PR #542, opened 2016-03-08
unset the max-width of the slider

jguffey:
Found a minor CSS issue. Seems bootstrap is giving class tooltop-inner a max width. Those with design-eyes might notice that larger labels won't be centered properly when given an explicit width.

With max-width: 200px
maxwidth
Without: 
maxwidthoff

Solution is simple, set max-width to none. Developers can still set a width or max-width later in their stylesheets if they wish to override.

### DIFF
--- a/src/less/rules.less
+++ b/src/less/rules.less
@@ -100,6 +100,7 @@
 	}
 	.tooltip-inner {
 		white-space: nowrap;
+		max-width: none;
 	}
 	.hide {
 		display: none;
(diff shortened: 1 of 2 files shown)


seiyria:
If you make a CSS change, please also make it to our SCSS.

jguffey:
Thanks, should be all set now.

rovolution:
@jguffey great eye! (it took me a while to notice)

PR looks solid. Thanks for your contribution!

jguffey:
Thanks, but of course, my designer noticed before I did!

PR #543, opened 2016-03-08
Recalculate size on relayout

MaZderMind:
If a slider is initially hidden (fior example inside a Bootstrap-Tab or any other `display:none` Container) one ought to call `.relayout`, so that the Slider can update the layout with the now available information (for example, most browsers report a width of 0 before the content is actually shown).

This leads to an issue, that is not yet handled in `.relayout()`:
 1. After the initialization of a hidden slider, `this._state.size` is 0
 2. Calling `.relayout()` after making it visible, internally calls `._layout()` which, among other things, recalculates the tick-lable size `labelSize`
 3. because `labelSize` is calculated based on `this._state.size`, all labels will also have a width of 0 and be displayed incorrect

This PR adds an additional call to `._resize()` into the body of the `.relayout()` method, which will update `this._state.size` before calling into `._layout()`.

The PR also contains an Example which shows how to use a Slider in an initially hidden Element (and reveals the issue, if the fix is not applied) as well as an automated test, which does the same.

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -921,6 +921,7 @@
 			},
 
 			relayout: function() {
+				this._resize();
 				this._layout();
 				return this;
 			},
--- a/test/specs/PublicMethodsSpec.js
+++ b/test/specs/PublicMethodsSpec.js
@@ -573,6 +573,38 @@ describe("Public Method Tests", function() {
     expect(tooltipMarginLeft).toBeGreaterThan(0);
   });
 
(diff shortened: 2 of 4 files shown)


seiyria:
Thanks! One thing you're missing is the before and after jsfiddles. We'd need to see those just so we can get a better grasp on what you're doing.

rovolution:
@MaZderMind any time to get those fiddles in here? would love to merge this PR but just want to confirm the before/after state

MaZderMind:
@rovolution sorry, havn't yet found the time to do it. maybe tomorrow. sry

rovolution:
@MaZderMind going to pull this down and run locally

rovolution:
looks great! merging in

PR #544, opened 2016-03-14
Fixing the vertical position slider. Fixes #308. Fixes #410.

### LINKED ISSUE #308: Fixed Position Vertical Slider
Hi, I am trying to have a fixed position vertical slider as part of the UI on my site. However there seems to be an issue once a user scrolls down the page. This seems similar to issue #286, however the fix doesn't seem to work in this situation.

You can recreate the issue by going to the demo page example 4, and in the element inspector add "position:fixed;top:0;" to it's wrapper element. Then when you scroll, the slider thumb goes out of sync by the amount that you scroll.

Any thoughts would be greatly appreciated,
thanks

### LINKED ISSUE #410: Slider handle affected by the viewport position
Hello, I have created a jsfiddle of the bug I am trying to fix and wondering if you could help.

Essentially I have the slider in a modal. 
If the viewport has scrolled down and the modal/slider is visible the slider handles seem relative to the viewport and do not work correctly - despite they look to be in the correct position.

https://jsfiddle.net/informa/a7c6s7zn/5/

**The Fiddle:**
+ The Buttons on the page open a bootstrap modal that contains a bootstrap slider - vertical.
+ The paragraph text on the page simulates a SCROLL on the body.
+ Clicking the green button opens the modal and the bootstrap slider works as expected.
+ Clicking the Red button, by scrolling past the text opens the modal but the bootstrap slider handle has been affected by the page scroll position and therefore doesn't work correctly.

tchoulihan:
Let me know if you need anything else for this.


seiyria:
Yes, we need a test and a before/after jsfiddle [as per our guidelines](https://github.com/seiyria/bootstrap-slider/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

rovolution:
I would like to see a fiddle of this version of the slider contained within a modal.

rovolution:
@tchoulihan any follow up on this?

tchoulihan:
I should be able to get to this tomorrow.

rovolution:
> I should be able to get to this tomorrow.

Awesome! thanks! really looking forward to this fix

tchoulihan:
[fiddle before](https://jsfiddle.net/serbiantx/xxbjmLks/1/)
[fiddle after](https://jsfiddle.net/Lf69ovhc/)

rovolution:
@tchoulihan you need to make this edit to the `src` file. You edited the distribution files, which are generated as part of our compile step

tchoulihan:
Okay, fixed in the src/ folder now.

rovolution:
breaking our test suite (also remove the contents of **dist** from the PR)

tchoulihan:
As in delete the entire dist folder in a commit? Why wouldn't the test suite do this itself, in order to regenerate it from source?

seiyria:
There's no reason for the test suite to do it. It's simply just something you shouldn't commit when making a PR.

rovolution:
@tchoulihan this is breaking our basic vertical slider example (Example #4 on the github.io page)



Could you look into having it such that your proposed fix can handle both these two bugs in addition to the case above? Thanks!

rovolution:
@tchoulihan I just pulled down your topic branch and the update is breaking Example #4 on our examples page. Is there a way you could come with a solution that addresses both your use case (slider within a modal) plus the use case for example 4 (slider within scrollable container)?

Issue #428, opened 2015-08-20
iOS touch vs scroll


jerrylow:
On iOS and possible Android (I have no way of testing). If I'm just scrolling but happened to tap on the screen where the slider is it changes the position of the toggle. We should potentially have a quick delay to see if the user's intention is to scroll the page or the slider.

seiyria:
Can you please make a demo or point one out where this behavior can be
viewed?

jerrylow:
Sure, here's a demo: http://codepen.io/anon/pen/LVKeQq

And a video: http://jerrylow.com/demo/bootstrap-slider/bootstrap-slider-demo.mp4

tobias-dev:
I'd like to see this too. Currently vertical scrolling on smartphones is basically not possible over slider elements. Maybe the slider should only work on tapping and horizontal scrolling/sliding on touch devices, but not on vertical scrolling (especially not prevent the vertical scrolling).

rovolution:
if you have ideas for how this could be implemented, please feel free to submit a PR. This would be a great idea to submit a PR for

jerrylow:
Alright, got a PR in there :)

rovolution:
fixed in 6.1.7

PR #551, opened 2016-03-31
Touch move conflicting with user trying to scroll fix

jerrylow:
Fixing issue #428 where scrolling on mobile devices are locked in the slider.

Here's the fixed version in demo: http://codepen.io/jerrylow/pen/wGrBVo

Also the previous issue with video demonstration:

Codepen: http://codepen.io/anon/pen/LVKeQq
Video: http://jerrylow.com/demo/bootstrap-slider/bootstrap-slider-demo.mp4

Changes:

I'm detecting the direction of the swipe before determining if they're trying to scroll or swipe on the slider. There's a threshold if they stay within 5px (+/-) of the slider direction also moving a distance greater than 15px. I did a bunch of tests with 5, 10, 15, 20, 25. Anything greater than 20 there was a bit of delay so it wasn't as responsive. Ten was ok but there's still an issue if the user is swiping slightly diagonal. Fifteen pixels seems to be the best balance to solve the issue on a basic level.

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1271,7 +1271,7 @@
 			},
 			_touchstart: function(ev) {
 				if (ev.changedTouches === undefined) {
-					this._trigger('slideStart');
+					this._mousedown(ev);
 					return;
 				}
 
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -540,6 +540,9 @@
 
 			this.touchCapable = 'ontouchstart' in window || (window.DocumentTouch && document instanceof window.DocumentTouch);
 
+			this.touchX = 0;
+			this.touchY = 0;
+
 			this.tooltip = this.sliderElem.querySelector('.tooltip-main');
 			this.tooltipInner = this.tooltip.querySelector('.tooltip-inner');
 
@@ -671,9 +674,13 @@
 			this.handle2.addEventListener("keydown", this.handle2Keydown, false);
 
 			this.mousedown = this._mousedown.bind(this);
+			this.touchstart = this._touchstart.bind(this);
+			this.touchmove = this._touchmove.bind(this);
+
 			if (this.touchCapable) {
 				// Bind touch handlers
-				this.sliderElem.addEventListener("touchstart", this.mousedown, false);
+				this.sliderElem.addEventListener("touchstart", this.touchstart, false);
+				this.sliderElem.addEventListener("touchmove", this.touchmove, false);
 			}
 			this.sliderElem.addEventListener("mousedown", this.mousedown, false);
 
@@ -957,7 +964,8 @@
 				if (this.hideTooltip) {
 					this.sliderElem.removeEventListener("mouseleave", this.hideTooltip, false);
 				}
-				this.sliderElem.removeEventListener("touchstart", this.mousedown, false);
+				this.sliderElem.removeEventListener("touchstart", this.touchstart, false);
+				this.sliderElem.removeEventListener("touchmove", this.touchmove, false);
 				this.sliderElem.removeEventListener("mousedown", this.mousedown, false);
 
 				// Remove window event listener
@@ -1261,6 +1269,16 @@
 
 				return true;
 			},
+			_touchstart: function(ev) {
+				if (ev.changedTouches === undefined) {
+					this._trigger('slideStart');
+					return;
+				}
+
+				var touch = ev.changedTouches[0];
+				this.touchX = touch.pageX;
+				this.touchY = touch.pageY;
+			},
 			_triggerFocusOnHandle: function(handleIdx) {
 				if(handleIdx === 0) {
 					this.handle1.focus();
@@ -1341,6 +1359,27 @@
 				this.setValue(val, true, true);
 
 				return false;
+			},
+			_touchmove: function(ev) {
+				if (ev.changedTouches === undefined) {
+					return;
+				}
+
+				var touch = ev.changedTouches[0];
+
+				var xDiff = touch.pageX - this.touchX;
+				var yDiff = touch.pageY - this.touchY;
+
+				if (!this._state.inDrag) {
+					// Vertical Slider
+					if (this.options.orientation === 'vertical' && (xDiff <= 5 && xDiff >= -5) && (yDiff >=15 || yDiff <= -15)) {
+						this._mousedown(ev);
+					}
+					// Horizontal slider.
+					else if ((yDiff <= 5 && yDiff >= -5) && (xDiff >= 15 || xDiff <= -15)) {
+						this._mousedown(ev);
+					}
+				}
 			},
 			_adjustPercentageForRangeSliders: function(percentage) {
 				if (this.options.range) {
--- a/test/specs/EventsSpec.js
+++ b/test/specs/EventsSpec.js
@@ -317,7 +317,7 @@ describe("Event Tests", function() {
         testSlider._mousemove(mouse);
         expect(numTimesFired).toEqual(1);
       });
-      
+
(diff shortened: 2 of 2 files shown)


rovolution:
@jerrylow failed our CI build.

Could you fix the linting problems and add the necessary automated tests to the TouchEvents section of our test suite which can be found [here](https://github.com/seiyria/bootstrap-slider/blob/6c0c6d69cbbb82e869c9b7127afc8d1e2838d58f/test/specs/EventsSpec.js#L120-L269)?

jerrylow:
Hey @rovolution might need some help around the testing. I've never used jasmine before but was poking around. This test I can't get to pass:

```
      it("slider should not bind multiple touchstart events after calling 'refresh'", function() {
        touch.initEvent("touchstart", true, true);
        flag = 0;

        testSlider.on('slideStart', function() {
          flag += 1;
        });
        testSlider.slider('refresh');
        $('.slider .slider-handle').get(0).dispatchEvent(touch);

        expect(flag).toEqual(1);
      });
```

All the unbinds should be happening but I can't really tell what is going on.

tobias-dev:
@jerrylow Thanks a lot for this great feature. I tried it myself and it works quite well.

One thing I noticed: [You mentioned above](https://github.com/seiyria/bootstrap-slider/pull/551#issue-145005078) the threshold of 5px to stay in the slider direction. However you seem to have implemented a 10px threshold instead of 5px, whereas 5px works much better for me. Basically when using 10px I would often accidentally slide while dragging the screen.

Examples:
- Dragging the screen with 10px threshold (accidentally changes the slider): http://screencast.com/t/RM8TkjPA
- Dragging the screen with 5px threshold (slider doesn't change): http://screencast.com/t/0ptbL8hWZwS

So from my point of view please use a threshold of 5px.

jerrylow:
Thanks @tobias-dev, moving it to 5px sounds like a good idea.

src/js/bootstrap-slider.js
```diff
@@ -1342,6 +1355,23 @@
 
 				return false;
 			},
+			_touchmove: function(ev) {
+				var touch = ev.changedTouches[0];
```
rovolution: The TravisCI server is breaking on the tests at this line. Here is the error: https://travis-ci.org/seiyria/bootstrap-slider/builds/120095399#L293-L295

image

Could you add a check in your code for when `ev.changedTouches[0]` is `undefined`?

src/js/bootstrap-slider.js
```diff
@@ -1261,6 +1269,11 @@
 
 				return true;
 			},
+			_touchstart: function(ev) {
+				var touch = ev.changedTouches[0];
```
rovolution: same feedback here as below (Could you add a check in your code for when `ev.changedTouches[0]` is `undefined`?)

rovolution:
@jerrylow that test is ensuring the `touchstart` listener is not bound twice if the `refresh()` method is invoked on the slider.

From glancing at our test suite, it looks like we already have tests to ensure the slider behaves as expected when touch events are triggered on it: https://github.com/seiyria/bootstrap-slider/blob/6c0c6d69cbbb82e869c9b7127afc8d1e2838d58f/test/specs/EventsSpec.js#L120-L269

As long as you can get our tests passing again, that should be fine.

jerrylow:
Ok thanks @rovolution I now understand why that test was failing.

src/js/bootstrap-slider.js
```diff
@@ -1261,6 +1269,16 @@
 
 				return true;
 			},
+			_touchstart: function(ev) {
+				if (ev.changedTouches === undefined) {
+					this._trigger('touchStart');
```
rovolution: could you have this trigger `slideStart` instead of `touchStart` as that is the event we have defined in our README?

rovolution: this also needs to emit the current value of the slider (similar to this line): https://github.com/jerrylow/bootstrap-slider/blob/Touch-Fix/src/js/bootstrap-slider.js#L1326

jerrylow: @rovolution I actually did that originally, but wasn't sure if that would cause any accidental event from happening that's bound to slideStart. Did you prefer slideStart?

test/specs/EventsSpec.js
```diff
@@ -218,7 +218,7 @@ describe("Event Tests", function() {
         touch.initEvent("touchstart", true, true);
         flag = 0;
 
-        testSlider.on('slideStart', function() {
+        testSlider.on('touchStart', function() {
```
rovolution: once you make the changes above, this should be renamed back to `slideStart`

src/js/bootstrap-slider.js
rovolution: from reading the code more closely, this method should probably call the `this._mousedown()`

src/js/bootstrap-slider.js
```diff
@@ -1342,6 +1360,27 @@
...
+
+				var touch = ev.changedTouches[0];
+
+				var xDiff = touch.pageX - this.touchX;
+				var yDiff = touch.pageY - this.touchY;
+
+				if (!this._state.inDrag) {
+					// Vertical Slider
+					if (this.options.orientation === 'vertical' && (xDiff <= 5 && xDiff >= -5) && (yDiff >=15 || yDiff <= -15)) {
+						this._mousedown(ev);
```
rovolution: this should be `this._mousemove()` since this is analogous to the `mousemove`

src/js/bootstrap-slider.js
```diff
@@ -1342,6 +1360,27 @@
...
+					}
+					// Horizontal slider.
+					else if ((yDiff <= 5 && yDiff >= -5) && (xDiff >= 15 || xDiff <= -15)) {
+						this._mousedown(ev);
```
rovolution: this should be `this._mousemove()` since this is analogous to the `mousemove`

src/js/bootstrap-slider.js
```diff
@@ -1261,6 +1269,16 @@
...
 			},
+			_touchstart: function(ev) {
+				if (ev.changedTouches === undefined) {
+					this._trigger('slideStart');
+					return;
+				}
+
+				var touch = ev.changedTouches[0];
+				this.touchX = touch.pageX;
+				this.touchY = touch.pageY;
```
rovolution: if you remove the function call from `line 1274` and call `this._mousedown(ev)` here, does that also work?

src/js/bootstrap-slider.js
jerrylow: I believe this touchmove state I added was just to validate that the user has reached a particular threshold. I see there's a bunch of initialization happening on mousedown that if we only call mousemove, won't happen. I was going with the idea that this touchmove just detects to initiate the actual mousedown using https://github.com/seiyria/bootstrap-slider/blob/master/src/js/bootstrap-slider.js#L1241 to work with the touchmove/mousemove event.

src/js/bootstrap-slider.js
jerrylow: Good call! It does and its actually the ultimate option as a fallback. This portion of the code won't work but that is ok, because the touchmove threshold check won't work either so this scenario everything will fallback to what it was before, which was functional at least.

src/js/bootstrap-slider.js
rovolution: > I was going with the idea that this touchmove just detects to initiate the actual mousedown using https://github.com/seiyria/bootstrap-slider/blob/master/src/js/bootstrap-slider.js#L1241 to work with the touchmove/mousemove event.

Good catch! Makes sense :+1:

rovolution:
alright, just tested your branch on my iPhone6 simulator and everything appears to work solid. going to merge

rovolution:
Thanks for the PR @jerrylow!

PR #553, opened 2016-04-05
Removed compass dependency. #529

### LINKED ISSUE #529: Removing compass dependency
Not everyone who uses `SASS`/`SCSS` uses ruby compass to transpile these to CSS.
`libSASS` is almost a drop-in replacement if you don't use image sprites or some of compass's special CSS3 libraries..

jcbvm:
Fixes #529 

It's a quick fix by just defining the mixins inline prefixed with `slider_` to prevent collisions. 

I think it's a good idea to separate the mixins and variables in partials and import them. If you want so, let me know.


rovolution:
> I think it's a good idea to separate the mixins and variables in partials and import them. If you want so, let me know.

If you could do that as well it would be much appreciated!

jcbvm:
Alright, should I restructure the folder to the same layout the less variant has? This will be a breaking change, because the filename will change from slider.scss to bootstrap-slider.scss. 

If so, maybe it's also a good idea to rename the variables (I think they are too global at the moment), something like the following to keep them in line with the less variant:

    $slider-line-height
    $slider-border-radius

rovolution:
> Alright, should I restructure the folder to the same layout the less variant has?
Go ahead and structure them this way in order to maintain consistency. we can do a major bump once we merge

jcbvm:
PR is ready to be merged. I tested the new build on a local project with success.

Changes in this PR:
- Removed compass dependency
- Created mixins for `border-radius`, `box-shadow` and `box-sizing` which were previously pulled in via compass.
- Renamed variables to `$slider-line-height` and `$slider-border-radius` to be in line with the less variant.
- Moved mixins and variables to separate files.
- Renamed `slider.scss` to `bootstrap-slider` and moved previouly content of `slider.scss` to a new file named `_rules.scss`, to be in line with the less variant (underscore is common sass syntax for files which need to be imported).
- Added `!default` suffix to variables (so they will only be set if not already set elsewhere, this gives users the ability to overwrite them).

rovolution:
This is awesome work @jcbvm...Thank you for this awesome PR! going to do a major bump as this might break existing builds for SASS users.

rovolution:
@jcbvm available at v7.0.0

jcbvm:
No problem, glad to help. Thanks for the release.

PR #562, opened 2016-04-13
Fix range slider swapping place on click

### LINKED ISSUE #547: Value[0] can be greater than value[1] in range slider
If you set both values equal in a range slider then click left of the value, value[1] will be set to the low value while value[0] is set to the high value. This shouldn't happen as value[0] should always be <= value[1]. This is easily reproducible in examples 2 or 16 on http://seiyria.com/bootstrap-slider/. As you click around it keeps this inverted order but the tiniest drag will snap back to the correct order.

value0 greater than value1

jerrylow:
Fix #547 issue where when range sliders overlap then clicking on the left moves the max handle to the left of the min handle.

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1222,6 +1222,7 @@
 					var diff1 = Math.abs(this._state.percentage[0] - percentage);
 					var diff2 = Math.abs(this._state.percentage[1] - percentage);
 					this._state.dragged = (diff1 < diff2) ? 0 : 1;
+					this._adjustPercentageForRangeSliders(percentage);
 				} else {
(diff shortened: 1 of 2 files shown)


seiyria:
Please add a test and a before/after slider (using the one present in the issue would be sufficient).

rovolution:
@jerrylow could you add an automated unit test for this? thanks!

jerrylow:
Yup, sorry I'll do that sometimes soon.

jerrylow:
@seiyria @rovolution - unit test added.

rovolution:
Just pulled down the branch and ran the demo page. Everything looks solid...Thanks again @jerrylow!

Issue #566, opened 2016-04-23
Q: How to adjust slider width to fill parent 100%?


attentiondmi:
Using CSS to adjust the slider's width is possible, but it results in the labels underneath keeping their old positions, and thus looking incorrect.

What's the best way to get the slider to expand its width to fit it's parent width 100%?

Thanks in advance!

AndreiRegiani:
Simply add this custom CSS, this class is already set to the input field, just override the value with !important

```
.slider {
    width: 100% !important;
}
```

or change this value from the library CSS itself

attentiondmi:
Thanks Andrei! I have tried that, but there's a problem: If I increase the
.slider width, then the .slider-tick-label width expands far outside its
container.

akitaylor:
The suggested solution works for me like a charm inside a bootstrap column, including the tick labels. The slider automatically relayouts itself when the width of the parent element changes.

seiyria:
The solution we suggest is the CSS one shown above. If tick labels are displacing, that is a bug and should be reported in a separate issue.

Issue #371, opened 2015-05-05
Passing slider any class via original input classes or via data-* attributes ?


baasha:
Hi,
is there any possibility to either set class of slider via data- attribute or (better) get classes from original input element ?

For example - I want in my bootstrap enabled project those two sliders:
red slider and yellow slider:

```html
<input class = "slider slider-danger">
<input class = "slider slider-warning">
```

When using original bootstrap-slider, it works like a charm, but this (**much better** ;) ) slider ignores original inputs classes.

Is there any workaround ?

Thanks.

seiyria:
It does not appear that we have support for that.

Bhoft:
this would be a one liner which could be added in the create Markup section like below.
```
this.sliderElem.className = "slider";
// get classes from the this.element and add them to the slider
this._addClass(this.sliderElem,this.element.className);
```

rovolution:
if you write automated tests, im all for it :)

Bhoft:
sadly i have no clue how to write this automated tests.
I did the change in my fork and also added a possibility that the css classes are automatically set if a tick is reached.

see my pull request
https://github.com/seiyria/bootstrap-slider/pull/395

rovolution:
Looks like an outdated issue...[since we do have similar functionality now with data-provide](https://github.com/seiyria/bootstrap-slider#using-bootstrap-slider-via-data-provide-api)

PR #579, opened 2016-05-25
Add variables to control slider width and height

jkrehm:
Adds LESS & SASS variables to control the width of the slider. It defaults to 210px (the currently hardcoded value) but can be overridden.

See #566 for a case that might be resolved with this change.

### DIFF
--- a/src/less/rules.less
+++ b/src/less/rules.less
@@ -3,7 +3,7 @@
 	vertical-align: middle;
 	position: relative;
 	&.slider-horizontal {
-		width: 210px;
+		width: @slider-horizontal-width;
 		height: @slider-line-height;
 		.slider-track {
 			height: (@slider-line-height/2);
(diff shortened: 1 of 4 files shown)


seiyria:
This seems simple and ok.

rovolution:
LGTM 👍  thanks for the PR @jkrehm!

j9850s:
why a specific width value? Why is the slider not 'reading' the default width of the form input control?

standaard bootstrap.
i expect a slider width of 100%; the class of the input must control the slider width, right?
<input type="text" class="form-control" id="sliderid" />
<script>
$("#sliderid").slider({});
</script>

seiyria:
@j9850s That is mostly irrelevant to this PR, as it was just creating an abstraction layer. If you want to, you can:

* make a PR for #371 (or chime in there)
* set the slider width via CSS

PR #584, opened 2016-06-08
Fix #481: Tick slider not snapping if clicked

### LINKED ISSUE #481: Tick slider not snapping to ticks if clicked
When dragging the tick slider works just fine, but when you click in between ticks, the value will be of that in between - its not snapping the way it is when you're dragging. And that is what I would think that should happen.

JSFiddle: http://jsfiddle.net/h3WDq/995/

jasonkonopacki:
Let me know if you need anything else. Issue #481 

Added _snapToValue(val) when calling _setDataVal and ._trigger('slideStop', in the _mouseup function.

### DIFF
--- a/dist/bootstrap-slider.js
+++ b/dist/bootstrap-slider.js
@@ -1376,8 +1376,8 @@ function _typeof(obj) { return obj && typeof Symbol !== "undefined" && obj.const
 				var val = this._calculateValue(true);
 
 				this._layout();
-				this._setDataVal(val);
-				this._trigger('slideStop', val);
+				this._setDataVal(this._snapToValue(val));
+				this._trigger('slideStop', this._snapToValue(val));
 
 				return false;
 			},


seiyria:
As per our [contribution guidelines](https://github.com/seiyria/bootstrap-slider/blob/master/.github/CONTRIBUTING.md) and [PR checklist](https://github.com/seiyria/bootstrap-slider/blob/master/.github/PULL_REQUEST_TEMPLATE.md), we'll need:

* unit tests (so this doesn't happen again)
* a before/after jsfiddle clearly demonstrating the issue

dist/bootstrap-slider.js
```diff
@@ -1376,8 +1376,8 @@ function _typeof(obj) { return obj && typeof Symbol !== "undefined" && obj.const
 				var val = this._calculateValue(true);
```
rovolution: this is a generated file. please make your change in the `/src` file.

Also...where does this `_snapToValue()` function come from? I dont see it defined anywhere in our source code

jasonkonopacki:
The issue appears to only happen with touch events and the fix is spotty. Will reopen when fixed.

dist/bootstrap-slider.js
jasonkonopacki: Wow, I totally missed on this one.

I was using dandrel's comment fix from #392 along with my new code.

Sorry about that guys, a bit new to github and lost track of myself making that change.

PR #598, opened 2016-07-13
refactor postpublish script into all-in-one release script

rovolution:
@seiyria 

I am attempting to refactor our release script a little bit.

One thing that always bothered me was the fact that we were always off by one patch bump when we published to npm.

So instead of having the version be bumped separately from the `gh-pages` build, `/dist` build, and `npm publish`, I figured we could just combine all of those into a script and then proxy it with the `npm run release` command. i tested it locally and it works great, but let me know what you think.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,6 @@ Want to use npm? `npm install bootstrap-slider`
 
 Want to get it from a CDN? https://cdnjs.com/libraries/bootstrap-slider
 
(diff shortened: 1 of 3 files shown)


seiyria:
This sounds good to me. I was never a fan of the multi-bump stuff that had to be done. Do you want me to merge it? It's labeled do not merge, so I'm unsure. Either way, it looks good to me. Assuming you end up doing releases this will work but I generally think platform npm scripts should be ... well, platform independent. But in this case I think it's fine enough.

rovolution:
ill go ahead and merge. thanks!

PR #608, opened 2016-07-27
Fix crash in Chrome

timmd909:
Chrome Version 47.0.2526.106 (64-bit) was throwing the following error:

	Uncaught SyntaxError: Block-scoped declarations (let, const,
	function, class) not yet supported outside strict mode

Firefox wasn't affected

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1656,7 +1656,7 @@
 
 		*********************************/
 		if($) {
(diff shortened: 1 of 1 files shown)


seiyria:
You should not be using src/bootstrap-slider.js, you should be using dist. Our src is compiled with babel.

timmd909:
Oops. I'll fix that. Thanks for getting back so quick.

Issue #585, opened 2016-06-15
Make it possible to remove slider knop


robsch:
I would appreciate if a slider knop could be removed. I have situations where no value should be set and the slider shouldn't show any value. I could image that `$slider.slider('setValue', null)` could be used to achieve this easily. Currently, I can just disable the slider and set the value to 0. But I would prefer if I can set it to no value.

Is this possible in another way already?

seiyria:
We talked about this a little in #346 and I believe another issue (can't find it atm) and the conclusion was that we will not do this because the blank bar is not helpful.

robsch:
Thank you so far. However, it's not exactly the same because I didn't think that the control would be activated in a scenario. It's just disabled and will remain disabled. It could be activated only with JS, not by the user.

My example: I show a picture and have a zoom slider. If the picture is not available the zoom slider should be still visible but disabled. I would prefer that the slider then doesn't show any value, instead of 0 or 100% or 0,75 or whatever.

Couldn't you just implement this? Nobody would be forced to use `null` as value. It's just an option for a special situation. And it's another feature.

seiyria:
This is feature creep IMO and I still hold that it's bad UX to allow this at all. You'll have to implement it in your own fork if you want it.

robsch:
Ok, thank you.

PR #612, opened 2016-08-09
Implementing a method to accept 'null'-values

lacrioque:
The default value of a slider is null.
This should also be updated in the README.md then.

Maybe this should be open for discussion.

Pull Requests
=============

- [ ] unit tests (we use [Jasmine 1.3](http://jasmine.github.io/1.3/introduction.html))
- [ ] JSFiddle (or an equivalent such as CodePen, Plunker, etc) with new feature or bug-fix
- [x] documentation updates to README file

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -762,7 +762,7 @@ const windowIsDefined = (typeof window === "object");
 				step: 1,
 				precision: 0,
 				orientation: 'horizontal',
-				value: 5,
+				value: null,
 				range: false,
 				selection: 'before',
 				tooltip: 'show',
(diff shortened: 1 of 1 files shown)


seiyria:
We've talked about this several times, and I am strictly against this because it is terrible UX.

seiyria:
See here: https://github.com/seiyria/bootstrap-slider/issues/585

lacrioque:
The reason behind the null value is not about UX in this case.
We have to create a method to securely see if the user has set a value on the slider.
So we do not want to set an initial value, if there is no value set by the user.
This should not interfere with the overall behaviour, just that the value is set after the slider was touched. Right now we are using a hidden field which is activated when the user touches the slider. 
This is not reliable, and so we are looking for other solutions. 
The 'null' was the shortest way for us.

seiyria:
Sorry, but we do have to consider UX even if you don't. Feel free to use your own fork, but we don't recommend it for UX reasons. I've ranted about this before so suffice to say this is not something that will be implemented.

PR #635, opened 2016-10-11
Corrected malformed string literal

george-viaud:
Pull Requests
=============
Can witness this failure/fix in IE11

Replaced apostrophes w/ single-quotes


seiyria:
These are for babel to preprocess the string. It's interpolated, not a literal.

george-viaud:
It causes IE11 to choke.

PR #636, opened 2016-10-11
Replaces const and let keywords with var useage for backwards compatibility (ie10 and prior, Android stock browsers etc)

george-viaud:
Pull Requests
=============

- [ ] unit tests (we use [Jasmine 1.3](http://jasmine.github.io/1.3/introduction.html))
- [ ] JSFiddle (or an equivalent such as CodePen, Plunker, etc) with new feature or bug-fix


rovolution:
@george-viaud we transpile our ES6 code via babel, so we do not recommend using the contents of the src directory. Please see the [dist](https://github.com/seiyria/bootstrap-slider/tree/master/dist) directory for the transpiled code

george-viaud:
Indeed many thanks for this - seiyria made me aware of this during a previous PR moments before this one but I had already submitted it.  I extend the apologies I offered him onto you as well.  Many thanks for the direction as and the introduction to Babel - it seems very interesting and useful - I'm sure I'll find great use of it.  -g

Issue #630, opened 2016-09-23
formatInvalidInputErrorMsg in _validateInputValue always throwing


ianchute:
Issues
=============
**NOTE:** Before filing a new issue, please search through the open issues list and check to see if an existing issue already exists that describes your problem. If so, leave a comment within that issue with the checklist items described below.

Please make sure you include the following in your issue report where appropriate:

- [ ] JSFiddle (or an equivalent such as CodePen, Plunker, etc) with the an example demonstrating the bug
- [ ] Instructions to replicate the bug in the Fiddle

Validation needs to be fixed to the following (to detect string numbers):

```
_validateInputValue: function _validateInputValue(val) {
				if (!isNaN(+val)) {
					return +val;
				} else if (Array.isArray(val)) {
					this._validateArray(val);
					return val;
				} else {
					throw new Error(ErrorMsgs.formatInvalidInputErrorMsg(val));
				}
			},
```

seiyria:
We would love a PR for this!

PR #642, opened 2016-10-26
Validation fix to allow for string numbers

snurby7:
# Pull Requests

- [x] unit tests (we use [Jasmine 1.3](http://jasmine.github.io/1.3/introduction.html))
- [x] JSFiddle (plunker [link ](https://plnkr.co/edit/TRhGSSQu9yCluH7xDHNC?p=preview) [toggle the two files, new is update]
- [x] Fix for #630 

### DIFF
--- a/test/specs/PublicMethodsSpec.js
+++ b/test/specs/PublicMethodsSpec.js
@@ -254,19 +254,6 @@ describe("Public Method Tests", function() {
         expect(sliderValue).toBe(5);
       });
 
-      it('should throw an error if the slider is given a value it cannot convert to an int', function(){
-        var valueToSet = "fail";
-        var originalSliderValue = testSlider.slider('getValue');
-
-        var settingValue = function() {
-          testSlider.slider('setValue', valueToSet);
-        };
-        expect(settingValue).toThrow(new Error( formatInvalidInputMsg(valueToSet) ));
-
-        var sliderValue = testSlider.slider('getValue');
-        expect(sliderValue).toBe(originalSliderValue);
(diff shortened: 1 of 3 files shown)


seiyria:
-1. IMO validate your data before passing it in.

snurby7:
@seiyria fair enough, (I do agree that it should be cleaned prior). You guys seemed to be open to the PR in #630  so I just wrapped it all up.

seiyria:
Sorry, some days I'm on autopilot. 

It's ultimately up to @rovolution but I think it's reasonable to require that the slider only operate on numbers.

test/specs/PublicMethodsSpec.js
```diff
@@ -246,6 +246,14 @@ describe("Public Method Tests", function() {
         expect(sliderValue).toBe(valueToSet);
       });
 
+      it("properly sets the value of the slider when given a string value", function(){
```
rovolution: could you also add a test for when you pass a string value that cannot be properly converted into a number?

For example: 

```js
testSlider.slider('setValue', "rolo");
```

snurby7:
@rovolution added an additional test and updated the readme. I feel as though the test is a little redundant as down in the file a little ways before the range slider tests there's a test "when an invalid value type is passed in" that passes in "a" as the value and throws and error, I can remove it if you feel the same way.

rovolution:
@snurby7 oops just saw that test! Thanks for pointing that out! Yea for sure go ahead and delete that test then since it would be redundant

Issue #640, opened 2016-10-20
Error when using destroy & reinitialize slider


maydemirx:
Hello,
i am trying to bind event 'slideStop' after destroy & reinitialize the slider then an error throwing 

> bootstrap-slider.js:970 Uncaught TypeError: Cannot read property 'push' of null

You can see the bug at https://jsfiddle.net/maydemirx/jcuo8ksp/

maydemirx:
By the way you checking the variable is undefined but after destroy the variable value is null.

rovolution:
found the issue i am making a fix

rovolution:
@maydemirx fix is published to 9.3.2

PR #644, opened 2016-10-30
Bug-fix: set this.eventToCallbackMap property values to undefined instead of null on destroy()

rovolution:
Fixes https://github.com/seiyria/bootstrap-slider/issues/640


seiyria:
Looks good to me. Maybe a semantics thing, but would it make more sense to `delete` it instead of waiting for it to get gc'd?

rovolution:
`delete` just removes the reference and marks it for GC on the subsequent GC pass. Won't make much of a difference other than removing the eventName key from the `this. eventToCallbackMap` object

rovolution:
@seiyria went ahead and refactored to use `delete`

rovolution:
@snurby7 merged and published to 9.4.0! thanks again!

PR #646, opened 2016-11-01
aria-valuetext as formatter value

mediaformat:
Pull Requests
=============

**Original issue** : Feature Request : aria-valuetext as formatter value
https://github.com/seiyria/bootstrap-slider/issues/641

**Example of feature in action**:
https://fiddle.jshell.net/bbw6z63n/4/

Issue #641

Issue #641, opened 2016-10-26
Feature Request : aria-valuetext as formatter value


mediaformat:
Using _formatter_ it is possible to modify the displayed value in the slider tooltip.

This is useful for many cases, such as irregularly spaced ticks, or unevenly stepped values, see :
#535

But also  when the rendered value cannot be accurately represented as a number; Such as week days, month names, etc.

Ex: https://jsfiddle.net/bbw6z63n/3/

Currently the bootstrap-slider is lacking a way to accessibly represent the _formatter_ value to assistive technologies.

**_aria-valuenow_** represents the _value_ of the slider, a helpful addition however would be to add an **_aria-valuetext_** attribute to represent the _formatter value_.

aria-valuetext : https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuetext_attribute

rovolution:
👍

rovolution:
published to 9.5.0. Thanks again for the high-quality PR @mediaformat!

mediaformat:
@rovolution Thanks for your patience

PR #657, opened 2016-11-22
tpl ex24 (range formatter)

mediaformat:
Example for 
#641  
#646  

Forgot to add an example...

### DIFF
--- a/tpl/index.tpl
+++ b/tpl/index.tpl
@@ -1454,7 +1454,7 @@ sliderTest.on('slide', function(slideEvt){
 		});
 		
 	   		/* Example 24 */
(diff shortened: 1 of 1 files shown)


rovolution:
We already have an example with a custom formatter function...[See Example #1 on our examples page](http://seiyria.com/bootstrap-slider/)

mediaformat:
Right, I thought you just [wanted an example](https://github.com/seiyria/bootstrap-slider/pull/646#discussion_r87324534)
The ARIA proprty will be shown whenever the formatter is used, so yeah example #1 is probably sufficient,

rovolution:
yea in this case since you already documented that the ARIA attribute is being set in the README, no need for an example.

closing this PR. but thanks again for taking the time to do this!

PR #669, opened 2016-12-08
Grab cursor on handle

madshostrup:


rovolution:
could you do the following?

**1)** add a screenshot or GIF to this PR as a comment that demonstrates this change?
**2)** Make this change in the `src` directory instead of the `dist` directory? `dist` is our compiled code

PR #680, opened 2017-01-03
Add passive option to touch event listeners.

crhallberg:
By adding a [passive option](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) to the `touchstart` and `touchmove` events we can let the browser know that we do not intend to ever cancel scroll events. I found no such cases in [`_touchstart`](https://github.com/crhallberg/bootstrap-slider/blob/2df8c2910ec0d63ddf2fb410430ec7fc24589785/src/js/bootstrap-slider.js#L1464) nor [`_touchmove`](https://github.com/crhallberg/bootstrap-slider/blob/2df8c2910ec0d63ddf2fb410430ec7fc24589785/src/js/bootstrap-slider.js#L1555). This will prevent any scroll lag that may be caused by this library.

[More on passive event listeners](https://developers.google.com/web/updates/2016/06/passive-event-listeners)

- [x] unit tests (we use [Jasmine 1.3](http://jasmine.github.io/1.3/introduction.html))
- [x] JSFiddle (no visible new features)
- [x] documentation updates to README file (N/A)
- [x] examples within [/tpl/index.tpl](https://github.com/seiyria/bootstrap-slider/blob/master/tpl/index.tpl) (for new options being added) (N/A)

There is also some trailing whitespace cleanup I can remove if you'd prefer.


seiyria:
I like this, but:

* please do not rebuild dist
* I think we should have a fallback for older browser (see [here](http://caniuse.com/#feat=passive-event-listener)), since an object is technically truthy, but the object format is not always supported.
* it's gotta pass travis

crhallberg:
Understood.

Is this a usual warning to get: ` WARNING: $.fn.slider namespace is already bound. Use the $.fn.bootstrapSlider namespace instead.`

seiyria:
If you have jquery UI slider, then I can see that happening. Otherwise, no.

crhallberg:
This is happening when I run the grunt test. Even on Travis. Does this look better? I found a way to fallback using a quick funtionality check.

seiyria:
I don't actually know how effectively that would work, to be perfectly honest.

crhallberg:
The detection is the same used by a few libraries (like [detect-it.js](https://github.com/rafrex/detect-passive-events/blob/master/src/index.js#L5-L9)) so I have confidence in it,

src/js/bootstrap-slider.js
```diff
@@ -741,9 +741,21 @@ const windowIsDefined = (typeof window === "object");
 			this.touchmove = this._touchmove.bind(this);
 
 			if (this.touchCapable) {
+				// Test for passive event support
+				var supportsPassive = false;
```
rovolution: could you declare this variable with `let` instead of `var` to scope to nearest enclosing block?

src/js/bootstrap-slider.js
```diff
@@ -741,9 +741,21 @@ const windowIsDefined = (typeof window === "object");
...
+				try {
+					var opts = Object.defineProperty({}, 'passive', {
```
rovolution: could you declare `opts` with `let` instead of `var` to ensure it is only scoped within the `try {}` block?

rovolution:
Great PR @crhallberg! In regards to the fallback check, [I saw that Paul Irish himself recommends a similar technique](https://github.com/Modernizr/Modernizr/issues/1894#issuecomment-198035521), so I am cool with that.

I left some feedback in the code so if you could incorporate that and ping me that would be much appreciated. Thanks!

crhallberg:
I resolved the conflict with the date change and converted my variables to `let` declarations.

rovolution:
Merged and published to v9.7.0. Thanks again for your contribution @crhallberg!

crhallberg:
Thanks for letting me!

Issue #682, opened 2017-01-04
Error compilation Sass


Schepotin:
v9.6.0

![image](https://cloud.githubusercontent.com/assets/6001723/21638186/eeac64dc-d272-11e6-944c-f568d38a5d25.png)

rovolution:
working to resolve this right now

rovolution:
should be resolved in v9.6.1

thanks for bringing to our attention so quickly! I added a TODO to add linting to the SASS/LESS files so we can be more proactive about catching these issues in our CI check phase

Shnoulle:
Oups ... sorry .... seems to be my pull request :blush:

PR #684, opened 2017-01-04
SASS/LESS linting

rovolution:
Adding in SASS/LESS linting so that we can quickly catch issues such [as this](https://github.com/seiyria/bootstrap-slider/issues/682) in the CI check step.

I tested this on the commit that triggered the above issue and it works like a charm!


seiyria:
Looks good to me!

PR #691, opened 2017-01-19
Prevent "'slider' of undefined" error

### LINKED ISSUE #587: TypeError: Cannot read property 'slider' of undefined
/dist/bootstrap-slider.js
Line: 1604

$.fn is undefined.

gutosanches:
Fixes #587

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1900,7 +1900,7 @@ const windowIsDefined = (typeof window === "object");
 			Attach to global namespace
 
 		*********************************/
-		if($) {
+		if($ && $.fn) {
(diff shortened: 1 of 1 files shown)


rovolution:
@gutosanches looks good!

Could you add a unit test to ensure that we can catch if this issue arises again?

It could be something as simple as temporarily poly-filling the `$.fn` method to `null`, instantiating an instance of a slider, making sure it does not throw a runtime error, and then restoring `$.fn` to whatever it originally was.

rovolution:
alright  I tried to write a good unit test for this, but its somewhat difficult to do well given how we currently test for errors that occur on script load.

rovolution:
merged and published to v9.7.1

PR #692, opened 2017-01-19
Fix #687: Accessibility on range slider with keyboard control

### LINKED ISSUE #687: Accessibility
Issues
=============

When we move a "two cursors slider" only with keyboard, interval is not correctly updated if one cursor passes over the other cursor (interval values should be inverted as they are when we cross cursors with a mouse).

jerrylow:
Fix issue #687 with keyboard control of range slider.

JSFiddle: https://jsfiddle.net/pgzp515m/

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1569,8 +1569,8 @@ const windowIsDefined = (typeof window === "object");
 				this._state.keyCtrl = handleIdx;
 				if (this.options.range) {
 					this._adjustPercentageForRangeSliders(percentage);
-          const val1 = (!this._state.keyCtrl) ? val : this._state.value[0];
-          const val2 = (this._state.keyCtrl) ? val : this._state.value[1];
+					const val1 = (!this._state.keyCtrl) ? val : this._state.value[0];
+					const val2 = (this._state.keyCtrl) ? val : this._state.value[1];
 					val = [ val1, val2];
 				}
 
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1565,9 +1565,12 @@ const windowIsDefined = (typeof window === "object");
 				}
 
 				var val = this._state.value[handleIdx] + dir * this.options.step;
+				var percentage = (val / this.options.max) * 100;
+				this._state.keyCtrl = handleIdx;
 				if (this.options.range) {
-					val = [ (!handleIdx) ? val : this._state.value[0],
-						    ( handleIdx) ? val : this._state.value[1]];
+					this._adjustPercentageForRangeSliders(percentage);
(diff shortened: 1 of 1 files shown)


jerrylow:
I am having issues running tests/phantomJS locally - not sure if anybody has any information on this. It's exactly this issue: https://github.com/gruntjs/grunt-lib-phantomjs/issues/89

src/js/bootstrap-slider.js
```diff
@@ -1565,9 +1565,12 @@ const windowIsDefined = (typeof window === "object");
 				}
 
 				var val = this._state.value[handleIdx] + dir * this.options.step;
+				var percentage = (val / this.options.max) * 100;
+				this._state.keyCtrl = handleIdx;
 				if (this.options.range) {
-					val = [ (!handleIdx) ? val : this._state.value[0],
-						    ( handleIdx) ? val : this._state.value[1]];
+					this._adjustPercentageForRangeSliders(percentage);
+					val = [ (!this._state.keyCtrl) ? val : this._state.value[0],
```
rovolution: could you break these array values into separate variables and then reference them within the array?

```js
const val1 = (!this._state.keyCtrl) ? val : this._state.value[0];
const val2 = this._state.keyCtrl) ? val : this._state.value[1];

val = [ val1, val2];
```

src/js/bootstrap-slider.js
```diff
@@ -1565,9 +1565,12 @@ const windowIsDefined = (typeof window === "object");
 				}
 
 				var val = this._state.value[handleIdx] + dir * this.options.step;
+				var percentage = (val / this.options.max) * 100;
```
rovolution: could you re-declare this as a `const`?

jerrylow:
@rovolution - updated. Yes let me know if you find out more. I've reinstall phantomJS in different ways and different places - but to be honest don't think that's the cause of the problem. This was not a problem before I upgraded to Sierra.

src/js/bootstrap-slider.js
```diff
@@ -1565,9 +1565,13 @@ const windowIsDefined = (typeof window === "object");
 				}
 
 				var val = this._state.value[handleIdx] + dir * this.options.step;
+				const percentage = (val / this.options.max) * 100;
+				this._state.keyCtrl = handleIdx;
 				if (this.options.range) {
-					val = [ (!handleIdx) ? val : this._state.value[0],
-						    ( handleIdx) ? val : this._state.value[1]];
+					this._adjustPercentageForRangeSliders(percentage);
+          const val1 = (!this._state.keyCtrl) ? val : this._state.value[0];
```
rovolution: could you fix the spacing? thanks!

jerrylow:
@rovolution done

rovolution:
@jerrylow merged and published to 9.7.2

rovolution:
thanks again for another great contribution!

PR #701, opened 2017-02-13
Added Prefix + Suffix to Slider Value + Zero / Middle track alignment

sifex:
Hi,

This PR adds two things
1. It adds the ability to add prefixes and [suffixes to the tooltip](http://i.imgur.com/iHBw347.png) (Which was required from the client)
2. Different alignment for the background track. ([Example with the track Middl'ed](http://i.imgur.com/iHBw347.png))

There hasn't been any unit tests added (which I'm happy to do, I'm just at work atm), and no JSFiddle, or templating. 

Happy to update those if you're interested in having these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,10 +180,12 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | orientation |	string | 'horizontal' |	set the orientation. Accepts 'vertical' or 'horizontal' |
 | value |	float,array |	5	| initial value. Use array to have a range slider. |
 | range |	bool |	false	| make range slider. Optional if initial value is an array. If initial value is scalar, max will be used for second value. |
-| selection |	string |	'before' |	selection placement. Accepts: 'before', 'after' or 'none'. In case of a range slider, the selection will be placed between the handles |
+| selection |	string |	'before' |	selection placement. Accepts: 'before', 'after', 'middle', 'zero' or 'none'. In case of a range slider, the selection will be placed between the handles |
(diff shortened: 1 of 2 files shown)


seiyria:
* your code needs to pass all of the CI
* we already have something to change the tooltip, use the `formatter` function
* I would need to see a before/after example for your #2 before I would consider it.

sifex:
@seiyria Cool, pretty sure codeclimate is failing because of something else, but I'll double check later. 

I'll hopefully setup a codepen tonight, if not just close this.

sifex:
Centered and Zeroed views example:
http://codepen.io/sifex/full/WRmbBO/

Let me know and I'll resend a PR with just this in it.

seiyria:
I'll be honest, I still don't really get it, can't you already do this? Why would you want this? I'm trying to understand the use cases here.

sifex:
screen shot 2017-02-15 at 11 00 12 am

This is one I prepared earlier. It's work I'm doing for a client atm. I didn't see anyway of accomplishing a centred track beforehand.

seiyria:
I see. If this is something that @rovolution agrees on then sure, but it will have to come with tests, example, and docs.

rovolution:
@sifex so I understand why you would want use case #2, but I'm not sure about the suffix/prefix feature. 

That functionality can just be accomplished via the formatter function...Just specify a custom formatter function where your prefix/suffix are prepended/appended

sifex:
@rovolution Yea, I'll submit another PR if you want me with just #2. Think it might be a good addition

rovolution:
@sifex that would be awesome thanks!

rovolution:
closing this for now based on discussion

PR #742, opened 2017-04-20
additional class support on 'rangeHighlights' option

jccode:
Hi seiyria, would you please accept the pull request? Thank you. 

screenshot


seiyria:
It looks like we need a unit test to make sure this works correctly. After that, I think it looks good.

jccode:
Yup, unit test are updated as well. 
I'm sorry that I uploaded the `dist` directory, which cause some "duplicate code" check error. Does it matter?

seiyria:
we'd prefer if it wasn't there, yes.

rovolution:
@jccode the dist directory is compiled code, which is generated from the code in the src directory. Please make your changes to the src directory code thanks!

jccode:
@rovolution I know. My code changes are already apply to src directory.  
Which makes me confused is the duplication is between `src/bootstrap-slider.js` and `dist/bootstrap-slider.js`. Would you please help to take a look at the detail check results, and  give some advice to fix this check issue?

seiyria:
Just don't commit `dist`. the difference is that `dist` is the built files, and `src` is the source code.

jccode:
@seiyria  I rollback the `dist` files. Now all checks have passed. Thank you.  :)

src/js/bootstrap-slider.js
```diff
@@ -472,12 +472,11 @@ const windowIsDefined = (typeof window === "object");
 
 				/* Create highlight range elements */
 				this.rangeHighlightElements = [];
-				if (Array.isArray(this.options.rangeHighlights) && this.options.rangeHighlights.length > 0) {
-					for (let j = 0; j < this.options.rangeHighlights.length; j++) {
-
+                var rangeHighlightsOpts = this.options.rangeHighlights;
```
rovolution: could you fix the spacing and declare this as a `const` instead of a `var` since the value is not being reassigned? thanks!

jccode:
Sorry that I wan't aware of the spacing issue. Thanks for your review.

src/js/bootstrap-slider.js
```diff
@@ -472,12 +472,11 @@ const windowIsDefined = (typeof window === "object");
...
-				if (Array.isArray(this.options.rangeHighlights) && this.options.rangeHighlights.length > 0) {
-					for (let j = 0; j < this.options.rangeHighlights.length; j++) {
-
+				const rangeHighlightsOpts = this.options.rangeHighlights;
+				if (Array.isArray(rangeHighlightsOpts) && rangeHighlightsOpts.length > 0) {
+					for (let j = 0; j < rangeHighlightsOpts.length; j++) {
 						var rangeHighlightElement = document.createElement("div");
-						rangeHighlightElement.className = "slider-rangeHighlight slider-selection";
-
+						rangeHighlightElement.className = "slider-rangeHighlight slider-selection " + (rangeHighlightsOpts[j].class || "");
```
rovolution: could you refactor this to use an ES6 template string instead? I think it will help make the code a bit cleaner:

```js
const customClassString = rangeHighlightsOpts[j].class || "";
rangeHighlightElement.className = `slider-rangeHighlight slider-selection ${customClassString}`;
```

jccode: WOW! Thanks for the great suggestion.

rovolution:
@jccode merged and published to 9.8.0. Thanks again for your contribution!

PR #758, opened 2017-06-30
Added link to Vue.js bindings

pimlie:
I have created Vue.js bindings so Vue.js users can also easily use the `bootstrap-slider` without too much efforts


seiyria:
thanks!

scottleblanc:
I get the following error when trying to load component in Vue 2:

Failed to mount component: template or render function not defined.

seiyria:
please open an issue on that tracker instead of discussing it here.

PR #759, opened 2017-07-04
Max value calculation wrong due to cutoff when max value is not a multiple of step

thaitzer:
gracefully handle max value when a step size is defined and the max v…alue does not fit the step size

### DIFF
--- a/test/specs/StepReachMaxValueSpec.js
+++ b/test/specs/StepReachMaxValueSpec.js
@@ -2,7 +2,6 @@ describe("TickMaxValueNotATickBehavior", function() {
 	var SLIDER_ID = "testSlider1";
 	var slider;
 	var options;
-  var cache = [];
 
 	describe('max value should be reached', function() {
 		beforeEach(function() {
@@ -38,22 +37,7 @@ describe("TickMaxValueNotATickBehavior", function() {
 	afterEach(function() { slider.destroy(); });
 
 	// helper functions
-	function clickTickAtIndexAndVerify(slider, tickIndex) {
-		var sliderLeft = slider.sliderElem.offsetLeft;
-		var tickLeft = slider.ticks[tickIndex].offsetLeft;
-		var handleHalfWidth = $('.slider-handle.round').width() / 2;
 
-		var offsetX = sliderLeft + tickLeft + handleHalfWidth;
-		var offsetY = slider.sliderElem.offsetTop;
-
-		var mouseEvent = getMouseDownEvent(offsetX, offsetY);
-
-		slider.mousedown(mouseEvent);
-		slider.mouseup();
-
-		var expectedValue = slider.options.ticks[tickIndex];
-		expect(slider.getValue()).toBe(expectedValue);
-	}
 
   function getMouseDownEvent(offsetXToClick, offsetYToClick) {
 		var args = [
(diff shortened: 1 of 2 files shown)


seiyria:
please add a unit test so this particular behavior doesn't happen anymore!

seiyria:
additionally, please make sure this passes our linter!

thaitzer:
I  have added to test and checked the linter.

However I have a problem with my test code. I did not find a useful way to get the outer right position for the slider right - i had to add 10px manually to the mouse-click. I do not know why this is the case but I figure you guys probably have a pointer for me :-). Then I will gladly fix the test and get rid of the "+10".

For the mouse-clicking I copied the method from another test - this is where codeclimate complains now about code duplication. What is your suggested way of fixing this?

test/specs/StepReachMaxValueSpec.js
```diff
@@ -0,0 +1,67 @@
...
+	var slider;
+	var options;
+
+	describe('max value should be reached', function() {
+		beforeEach(function() {
+			options = {
+				min: 39,
+				max: 1310,
+				step: 5,
+        scale:"logarithmic",
```
rovolution: test is solid! could you fix the spacing on this line?

test/specs/StepReachMaxValueSpec.js
```diff
@@ -0,0 +1,67 @@
...
+				value: 44
+			};
+
+			slider = new Slider(document.getElementById(SLIDER_ID), options);
+		});
+
+		it("Value should contain max value when slider is moved to outer right position", function() {
+
+			var sliderLeft = slider.sliderElem.offsetLeft;
+      var offsetY = slider.sliderElem.offsetTop;
```
rovolution: spacing

test/specs/StepReachMaxValueSpec.js
```diff
@@ -0,0 +1,67 @@
...
+			var offsetX = sliderLeft + slider.sliderElem.clientWidth+10;
+
+      var expectedValue = slider.options.max;
```
rovolution: spacing

rovolution:
> However I have a problem with my test code. I did not find a useful way to get the outer right position for the slider right - i had to add 10px manually to the mouse-click. I do not know why this is the case but I figure you guys probably have a pointer for me :-). Then I will gladly fix the test and get rid of the "+10".

as long as you feel the +10 value is roughly accurate, that is fine. maybe include a comment in the test about why you are using this value

> For the mouse-clicking I copied the method from another test - this is where codeclimate complains now about code duplication. What is your suggested way of fixing this?

Don't worry about this. I have been pondering completely removing the codeclimate check or just limiting it to the source directory.

thaitzer:
I clean up the test code, added the comment and fixed the intendation.

rovolution:
@thaitzer just pulled down your branch to run the examples page and it works great!

rovolution:
@thaitzer btw, are there any existing issues from our issues list that this fix addresses?

I just want to ensure those get listed in our changelog.

https://github.com/seiyria/bootstrap-slider/issues

rovolution:
@thaitzer published to v9.8.1: https://github.com/seiyria/bootstrap-slider/blob/master/CHANGELOG.md#981--2017-07-06

thaitzer:
@rovolution I did not check the issues before writing this mini-patch. So the answer is: not that I know of.

While I am writing this: I never bothered to check if the same thing happens on the linear scale - I might be a good idea to do that.

rovolution:
> While I am writing this: I never bothered to check if the same thing happens on the linear scale - I might be a good idea to do that.

if you find that this is also an issue, we would definitely appreciate another PR!

Issue #761, opened 2017-07-10
shiftKey and CtrlKey missing in Slide event


drhouse82:
Hi,
I would like to detect whether AltKey or CtrlKey is pressed during a Slide Event.
However, it appears to me, that the original mouse event (containing this information) underlyling the SlideEvent is not included in the latter.

Would it be possible to pass the original mouse event as property of the slideEvent?

seiyria:
If this functionality interests you, please submit a PR.

PR #763, opened 2017-07-10
added originalEvent to sliderCallbacks

drhouse82:
PR for issue #761 .

The Original Mouse or Keyboard event is added to events slide, slideStop, slideStart.

jsfiddle: https://jsfiddle.net/2Lpbdgha/2/ 

* this._trigger now has an additional optional argument originalEvt to forward the original mouse or keyboardEvent to the Slider-Event.
* setValue also has an additional optional argument originalEvt.
* in non-jQuery mode, the original event is passed to the callback as additional argument.
* in jQuery mode, the original event is integrated as property "originalEvent" into the jQuery event.


src/js/bootstrap-slider.js
```diff
@@ -1779,26 +1779,28 @@ const windowIsDefined = (typeof window === "object");
...
+					this._triggerJQueryEvent(evt, val, originalEvt);
 				}
 			},
-			_triggerJQueryEvent: function(evt, val) {
+			_triggerJQueryEvent: function(evt, val, originalEvt) {
 				var eventData = {
 					type: evt,
-					value: val
+					value: val,
+                    originalEvent: originalEvt
```
seiyria: this is indented a space too far

seiyria:
I was going to add comments to every line but basically everything is indented a space too far. also, please add a test that verifies originalEvt exists and has some of the expected values that you find useful.

drhouse82:
OK. 
In my webstorm SDK indentation looks good... strange.
Also when I watch the commit's changes on the github page, the old and new versions seem to have the same indentation...

drhouse82:
webstorm IDE...

drhouse82:
Now checking existence of originalEvt in _trigger() and setValue().
_triggerJQueryEvent is only called from _trigger, so it needs no additional check (in my opinion).
It will also just includ an undefined field in worst case. What do you think?

drhouse82:
should I add some comment??

src/js/bootstrap-slider.js
```diff
@@ -898,10 +898,14 @@ const windowIsDefined = (typeof window === "object");
 				}
 			},
 
-			setValue: function(val, triggerSlideEvent, triggerChangeEvent) {
+			setValue: function(val, triggerSlideEvent, triggerChangeEvent, originalEvt) {
 				if (!val) {
 					val = 0;
 				}
+                if (!originalEvt) {
```
rovolution: fix spacing

src/js/bootstrap-slider.js
```diff
@@ -898,10 +898,14 @@ const windowIsDefined = (typeof window === "object");
 				}
 			},
 
-			setValue: function(val, triggerSlideEvent, triggerChangeEvent) {
+			setValue: function(val, triggerSlideEvent, triggerChangeEvent, originalEvt) {
```
rovolution: could you update the README to reflect this new change to the method signature?

https://github.com/seiyria/bootstrap-slider#functions

src/js/bootstrap-slider.js
```diff
@@ -1779,26 +1783,30 @@ const windowIsDefined = (typeof window === "object");
 				this.element.setAttribute('value', val);
 				this.element.value = val;
 			},
-			_trigger: function(evt, val) {
+			_trigger: function(evt, val, originalEvt) {
 				val = (val || val === 0) ? val : undefined;
+                if (!originalEvt) {
```
rovolution: fix spacing

src/js/bootstrap-slider.js
rovolution: fix spacing

drhouse82:
I've updated signature and fixed spacings.

For the unit test, I likely need some guidance... I've never done this before.
I tried to come up with a test...
I can trigger a slide event programatically, Yet, this doesnt lead to the original append being transmitted to the callback. I would have to trigger a mouse event on the slider-handle / slider-track that makes the slider trigger a slide event by itself...
however, I failed in doing so...

Any advice?

drhouse82:
sorry for the typo: append => event

rovolution:
> For the unit test, I likely need some guidance... I've never done this before.

1) Within a test block (defined by the `it` declaration), you need to initialize a slider and bind it to an event, such as `slide`). [See this test as an example](https://github.com/seiyria/bootstrap-slider/blob/master/test/specs/EventsSpec.js#L34-L36)
2) After binding the event listener to the slider, you need to:
* create a fake mouse event in JS
* apply it to the slider
[See this example](https://github.com/seiyria/bootstrap-slider/blob/f7219c3e704239b17678b7cd12a953e8ab692465/test/specs/StepReachMaxValueSpec.js#L24-L25)
3) After triggering the event, you need to:
* assert that the event listener was triggered
* Make sure the originalEvent is contained within the event object in the event listener callback.

Since the `originalEvent` only applies to jQuery based events, you will probably have to make sure you bind to the slider using the JQuery event system (the example I linked to in step 1 uses this event system, so if you follow that pattern you should be good).

drhouse82:
hi!
I updated the readme.
While I think, I could handle the test

drhouse82:
I adjusted the Readme.

Moreover, now, no matter wheter initialized with jquery or without jquery, the original mouse event is passed as second argument to the slide, slidestart, slidestop callbacks. (Previously, in jquery mode it was integrated as property inside the slide-event).
I managed to test the existance of these transmitted events for the non-jquery-initialized sliders by simulating mousedown, mousemove, mouseup.

For the jquery-initialized slider, I can only stimulate the slider to fire the slideStart event, not slide or slideStop... Here is a jsfiddle that does essentially the same:
https://jsfiddle.net/2Lpbdgha/25/ 
is_jquery variable in the beginning defines the mode of initialization, console output shows which events are fired and if transmitted events are contained.

Can you help me with the jquery mousemove event? I tried to dispatch the event to different elements... Actually I also checked out that the transmitted mouse event from a REAL mouse move is targeted to div.slider-handle.round. However, dispatching a synthetic mouse event there has no effect.

For the codeclimate test: it doesnt like my polyfill for mouseEvent (which is not supported by travis PhantomJS browser). Moreover, I don't see how to avoid the similar code in the slider callback-bindings... I believe, in non-jq-mode, I cannot decide inside he callback what the event type is.

README.md
```diff
@@ -225,9 +225,9 @@ Events
 ======
 | Event | Description | Value |
 | ----- | ----------- | ----- |
-| slide | This event fires when the slider is dragged | The new slider value |
-| slideStart | This event fires when dragging starts | The new slider value |
-| slideStop | This event fires when the dragging stops or has been clicked on | The new slider value |
+| slide | This event fires when the slider is dragged | The new slider value, the original mouse or keyboard event |
+| slideStart | This event fires when dragging starts | The new slider value, the original mouse or keyboard event |
+| slideStop | This event fires when the dragging stops or has been clicked on | The new slider value, the original mouse or keyboard event |
```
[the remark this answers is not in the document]
drhouse82: OK... requested changes done, all tests passed.

drhouse82:
Dear reviewers,
I feel I have incorporated all feedback and pass the CI tests. Is there anything else I need to do?

test/specs/originalEvtSpec.js
```diff
@@ -0,0 +1,99 @@
...
+            shiftKey: false,
+            altKey: false
+        };
+        var event = new MouseEvent(type, args);
+        return event;
+    }
+
+    // MouseEvent constructor  polyfill
+    function polyfill(window) {
+        var MouseEvent3 = function (eventType, params) {
```
rovolution: sorry for the late reply @drhouse82 but this all looks awesome!

One last request...could you move this polyfill to a separate file, add it to the package.json, and then include it in the jasmine test page via the grunt file?

See the `phantom_bind_polyfill.js` as an example:

* https://github.com/seiyria/bootstrap-slider/blob/master/test/phantom_bind_polyfill.js
* `package.json reference:` https://github.com/seiyria/bootstrap-slider/blob/286f98389971bc521364778205e2d704a4cefb3d/package.json#L70
* `Gruntfile reference:` https://github.com/seiyria/bootstrap-slider/blob/458d7ab6c4e776a9c0bfeff7e526591ac4191b82/Gruntfile.js#L125

rovolution:
@drhouse82 closing b/c this is old. please rebase against current `master` if you are interested in it being re-reviewed and merged. thanks!

Issue #778, opened 2017-09-18
Can we get color variables (utilizing the brand-primary variable) in the source SCSS/LESS files?


jmarikle:
I'm using bootstrap-sass on my project and I noticed that after changing the `$brand-primary` variable in bootstrap, it didn't change the color for the slider handle.  Can we get the slider to use the bootstrap color variables and maybe get the other hardcoded colors like the grays into variables too?  I can create a pull request if you need.

seiyria:
something like this would definitely require a pull request, however we do not have a hard dependency on bootstrap, just a soft one, so it might just be easier to change the color of the slider to match.

PR #779, opened 2017-09-18
switched to using variables for the colors

jmarikle:
Pull Request Referencing #778 
=============

Requesting pull for branch that uses variables in place of hardcoded values for the control colors.  Additionally, brand-primary is used.  bootstrap variables file is already a dependency of bootstrap-slider.less by use of `@line-height-computed` in variables.less, so that just uses the `@brand-primary` variable.  For the SASS file, it either uses the `$brand-primary` variable if it exists or it uses a default of `#0480be`.

Feel free to tweak the colors as needed.  Otherwise, if they need adjusted, I can make the changes in another commit too.

### DIFF
--- a/src/less/rules.less
+++ b/src/less/rules.less
@@ -27,7 +27,7 @@
 				border-width: 0 (@slider-line-height/2) (@slider-line-height/2) (@slider-line-height/2);
 				width: 0;
 				height: 0;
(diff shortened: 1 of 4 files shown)


rovolution:
@jmarikle changes merged and published to v9.9.0

PR #788, opened 2017-10-20
Remove bootstrap from dependencies and make use of node_modules bootstrap

patchman-sebastiaan:
The dependencies (bootstrap and highlightjs) are used in the examples. `bootstrap.min.css` tries to include the glyphicons font. Glyphicons are not used in the examples, however, it will result in 404's because of the fact that there are broken urls to the glyphicon files.

Our project recursively scans all css files in our project and looks for broken URLs. Bootstrap-slider is not usable for us because it will fail when it tries to follow links in `bootstrap.min.css`. Adding the fonts will solve this problem.


seiyria:
I'm not inclined to merge this as we don't appear to have a hard dependency on bootstrap - see [here](https://github.com/seiyria/bootstrap-slider/tree/master/src) for our styles.

patchman-sebastiaan:
I can agree that bootstrap is not a hard dependency. However, at this moment the library is referring to files that doesn't exist. I think that's wrong and as I said, it makes the library unusable for us and most probably also for other people that are using this package in Django projects. If you're not inclined to merge this, I would opt for removing the bootstrap file from the library (and maybe include it with bower when building the examples).

seiyria:
In that case I would indeed rather remove the offending css files. I don't
know the specifics of why it's included or how your system is finding it
though.

patchman-sebastiaan:
In that case I've changed the PR to remove `dependencies/css/bootstrap.min.css` and make use of the bootstrap in node_modules.

seiyria:
@rovolution what do you think? Will that break anything?

rovolution:
my only concern is that it this may break the current setup for deploying our examples page.

can this be made in a way that does not break the gh-pages branch? perhaps updating our build process to move the copy of bootstrap slider from our `node_modules` directory to the `gh-pages` branch?

patchman-sebastiaan:
Ah heck, didn't think of the gh-pages branch. Made some changes to incorporate that branch as well, and make sure nothing breaks there.

rovolution:
@patchman-sebastiaan changes looks solid! did you at least test the docs page locally? just run `grunt dev` and then navigate to `http://localhost:9000/` on your browser and make sure it all looks the same as the page being served off our `gh-pages` branch: http://seiyria.com/bootstrap-slider/

patchman-sebastiaan:
Yes, I did tested the docs page locally, and it works just like before.

Op 26 okt. 2017 06:09 schreef "Rohit Kalkur" <notifications@github.com>:

@patchman-sebastiaan <https://github.com/patchman-sebastiaan> changes looks
solid! did you at least test the docs page locally? just run grunt dev and
then navigate to http://localhost:9000/ on your browser and make sure it
all looks the same as the page being served off our gh-pages branch:
http://seiyria.com/bootstrap-slider/

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/seiyria/bootstrap-slider/pull/788#issuecomment-339545650>,
or mute the thread

scripts/update-gh-pages.sh
```diff
@@ -8,17 +8,19 @@ echo "..."
 grunt build-gh-pages
 # Create temporary copy of index file
 cp index.html index-temp.html
+cp node_modules/bootstrap/dist/css/bootstrap.min.css temp/bootstrap.min.css
```
rovolution: could you remove this step to be after line 15? i think it makes more sense there

rovolution:
@patchman-sebastiaan left one final piece of feedback. otherwise it looks awesome and we will definitely merge

patchman-sebastiaan:
Your requested change makes indeed more sense. Changed it and can confirm it works for gh-pages (https://patchman-sebastiaan.github.io/bootstrap-slider/).

rovolution:
merged and released to v9.10.0. Thanks again for your contribution!

PR #789, opened 2017-10-26
Fix sass mixin usage

armpogart:
Minor fix for sass, the mixin was used incorrectly.

### DIFF
--- a/src/sass/_rules.scss
+++ b/src/sass/_rules.scss
@@ -181,7 +181,7 @@
 
   position: absolute;
   background: transparent;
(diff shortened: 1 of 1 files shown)


seiyria:
Looks simple enough for me.

armpogart:
Another fix for the linter (mixins should come before declaration) to pass CI build.

rovolution:
merged and released to v9.9.1. thanks again for your contribution @armpogart!

armpogart:
np, I'm glad I could fix even such a small bug on the plugin that I use almost on all my projects. Thanks for your great product!

PR #798, opened 2017-11-02
Use `transform: translate()` instead of margin-* for the tooltips. Fix #483

### LINKED ISSUE #483: Chrome/Split tooltip on range slider not aligned
I'm using Chrome 47 for Windows and a range slider with split tooltips which are also set to always show. Until I click on the slider the tooltips are always misaligned, then they jump to the correct place. This is within a bootstrap modal. I've changed the tooltip CSS but then the tooltips will jump on click and be incorrect afterward. 

image

The code I'm using is pretty straight up -  `function gradeFilter() {
                    var slider = $("#gradeFilter").bootstrapSlider({
                        tooltip: 'always',
                        tooltip_split: true
                    });`

wismill:
Fix #483 using CSS property `transform` instead of `margin-*` for the tooltips.

According to [caniuse.com](http://caniuse.com/#search=transform), this property has a very good support in the browsers.

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -651,9 +651,6 @@ const windowIsDefined = (typeof window === "object");
 					this._removeProperty(tooltip, 'left');
 					this._removeProperty(tooltip, 'right');
 					this._removeProperty(tooltip, 'top');
-					this._removeProperty(tooltip, 'margin-left');
-					this._removeProperty(tooltip, 'margin-right');
-					this._removeProperty(tooltip, 'margin-top');
 
 					this._removeClass(tooltip, 'right');
 					this._removeClass(tooltip, 'left');
@@ -1153,11 +1150,6 @@ const windowIsDefined = (typeof window === "object");
 				this._setText(this.tooltipInner, formattedTooltipVal);
 
 				this.tooltip.style[this.stylePos] = `${positionPercentages[0]}%`;
-				if (this.options.orientation === 'vertical') {
-					this._css(this.tooltip, `margin-${this.stylePos}`, `${-this.tooltip.offsetHeight / 2}px`);
-				} else {
-					this._css(this.tooltip, `margin-${this.stylePos}`, `${-this.tooltip.offsetWidth / 2}px`);
-				}
 
 				function getPositionPercentages(state, reversed){
 					if (reversed) {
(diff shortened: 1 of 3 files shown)


seiyria:
this would need to pass our full CI suite before we could merge it. You'll also need to follow the contribution guidelines: https://github.com/seiyria/bootstrap-slider/blob/master/.github/CONTRIBUTING.md

You need to add a unit test to replace the one that was taken away, and I would like to see a jsfiddle of this.

wismill:
Well, it passes the full CI as I did not modify any of the lines reported by `codeclimate`.
I do follow your contribution guidelines. Please indicate what is not respected.
The unit test does not apply anymore, as the margin of the margins of the tooltips are never modified.

seiyria:
- It will need to pass code climate, regardless of whether you touched it or not
- You should add a jsfiddle so we can see it in action
- You should add a _new_ unit test in place of the old one.

wismill:
The issue detected by codeclimate is a duplication between javascript files in `src` and `dist`. How can I solve this?

seiyria:
I'm not sure why it would come up then. Maybe @rovolution knows why?

Either way though - the rest applies.

wismill:
JSFiddle:
- Original (see #483) http://jsfiddle.net/cwssL37d/
- With updated bootstrap-slider : http://jsfiddle.net/5waxge9b/1/

wismill:
Re-added the test, modified.

rovolution:
My one concern is that since [we state in our docs](https://github.com/seiyria/bootstrap-slider#supported-browsers) that we support IE9 and above, this will break that guarantee.

That being said, I would be ok with merging/releasing this, but would want to do so under a major version bump.

wismill:
It [is supported by IE 9](https://caniuse.com/#search=translate) using prefix `-ms-`. But this will add more code. Or all of this could be managed in CSS by using a class rather than a fixed CSS property. I'll check that.

wismill:
I have simplified the code and added support for IE 9. In fact, your code already uses CSS `transform` for `.slider-handle.triangle` and without compatibility.

rovolution:
> The issue detected by codeclimate is a duplication between javascript files in src and dist. How can I solve this?

Yea I agree this is a red-herring error. I'm going to try to find a way to ignore any changes to the dist directory for code climate checks

rovolution:
merged and released to v10.0.0 (published as a breaking change just incase there any inadvertent regressions). Thanks again for your contribution!

Issue #689, opened 2017-01-16
Bootstrap 4 Compatibility


jeremylynch:
Tooltips are not working in bootstrap v4 alpha 6. 

The HTML for tooltips has changed: 

**Bootstrap 3 html:** 
```html
<div class="tooltip top" role="tooltip">
  <div class="tooltip-arrow"></div>
  <div class="tooltip-inner">Tooltip on the left</div>
</div>
```

**Bootstrap 4 html:** 
```html
<div class="tooltip tooltip-top" role="tooltip">
  <div class="tooltip-inner">Tooltip on the top</div>
</div>
```

## We have three things to change: 
1. `.top` is now `.tooltip-top`
2. `.tooltip-arrow` has been removed
3. Change `.in` to `.show`

After making these changes my tooltip position was still wrong. I fixed this with: 

```css
.tooltip.tooltip-main {
  margin-top: -40px;
}
```

Shnoulle:
Maybe, wait for boostrap 4 beta version. alpha is unstable : maybe in was added in some day.

seiyria:
Would be happy to have a PR that adds these classes to the LESS/SCSS files!

mrjlynch:
@seiyria do you think we should just take the bootstrap 3 tooltip SCSS that has been removed in v4 and add it to the bootstrap-slider SCSS?

seiyria:
It'd probably just be a matter of copying and pasting, yes. We don't want any hard reliance on bootstrap.

natcohen:
any chance this can be resolved quickly so we can use bootstrap v4?

seiyria:
Sure, if you submit a PR for it.

taha333taha:
just add : 
.slider.slider-horizontal:hover,
.slider.slider-vertical:hover {
	.tooltip.tooltip-main {
	  opacity: 1 !important;
	}
}

str:
@taha333taha in my case, the tooltip was too big. This is the fix I added to my SCSS:

```SCSS
 .slider.slider-horizontal:hover, .slider.slider-vertical:hover {
    .tooltip.tooltip-main {
        opacity: 0.7 !important;
    }
    .tooltip-inner {
        font-size: 90%;
        width: auto;
    }
}
```

str:
Also, the initial value is showing way to the left:

![slider-left](https://user-images.githubusercontent.com/123817/26934780-9501e48c-4c27-11e7-8a71-1b8b35ed2bd4.png)

[...]

PR #819, opened 2018-02-22
introduce bs_version option for bs4 compatibility

madflow:
- [x] JSFiddle (or an equivalent such as CodePen, Plunker, etc) or screenshot/GIF with new feature or bug-fix
- [x] documentation updates to README file

----

This is a first stab at Bootstrap 4 compatibilty. The idea here: Introduce a `bs_version` (`bs3` is default) (`data-slider-bs-version="bs4|bs3"`) option in order to simply support both upstream versions (3+4) in the same branch. It simply deals with new `Tooltip` css classes introduced in BS4. It is somewhat hacky :S 

The tooltips work - but they seem to need some more positioning fixes...

Fixes: https://github.com/seiyria/bootstrap-slider/issues/689
Plunkr with bs4 hard-coded as default: https://plnkr.co/edit/6fyg2tpFhNdKHQdgZ757

I have not added an example to the `tpl/index.tpl` because I am unsure how it would be best to:

* Install both versions of Boostrap as dev dependencies and have a sane build pipeline
* Deal with no "Less" files in the Bootstrap 4 package

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | focus | bool | false | Focus the appropriate slider handle after a value change. |
 | labelledby | string,array | null | ARIA labels for the slider handle's, Use array for multiple values in a range slider. |
 | rangeHighlights | array | [] | Defines a range array that you want to highlight, for example: [{'start':val1, 'end': val2, 'class': 'optionalAdditionalClassName'}]. |
+| bsVersion | string | 'bs3' | Switch between Boostrap 3 ('bs3') and Bootstrap 4 ('bs4') |
 
 Functions
 =========
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -647,14 +647,15 @@ const windowIsDefined = (typeof window === "object");
 				}, this);
 
 				// Undo inline styles and classes on tooltips
+				const positionClasses = this._getTooltipPositionClassMap();
 				[this.tooltip, this.tooltip_min, this.tooltip_max].forEach(function(tooltip) {
 					this._removeProperty(tooltip, 'left');
 					this._removeProperty(tooltip, 'right');
 					this._removeProperty(tooltip, 'top');
 
-					this._removeClass(tooltip, 'right');
-					this._removeClass(tooltip, 'left');
-					this._removeClass(tooltip, 'top');
+					this._removeClass(tooltip, positionClasses.right);
+					this._removeClass(tooltip, positionClasses.left);
+					this._removeClass(tooltip, positionClasses.top);
 				}, this);
 			}
 
@@ -879,7 +880,8 @@ const windowIsDefined = (typeof window === "object");
 				focus: false,
 				tooltip_position: null,
 				labelledby: null,
-				rangeHighlights: []
+				rangeHighlights: [],
+				bsVersion: 'bs3'
 			},
 
 			getElement: function() {
@@ -1124,23 +1126,28 @@ const windowIsDefined = (typeof window === "object");
 					delete this.eventToCallbackMap[eventName];
 				}
 			},
+			_getTooltipShowClass: function() {
+				return (this.options.bsVersion === 'bs4') ? 'show' : 'in';	
+			},
 			_showTooltip: function() {
+				const showClass = this._getTooltipShowClass();
 				if (this.options.tooltip_split === false ){
-					this._addClass(this.tooltip, 'in');
+					this._addClass(this.tooltip, showClass);
 					this.tooltip_min.style.display = 'none';
 					this.tooltip_max.style.display = 'none';
 			    } else {
-					this._addClass(this.tooltip_min, 'in');
-					this._addClass(this.tooltip_max, 'in');
+					this._addClass(this.tooltip_min, showClass);
+					this._addClass(this.tooltip_max, showClass);
 					this.tooltip.style.display = 'none';
 				}
 				this._state.over = true;
 			},
 			_hideTooltip: function() {
+				const showClass = this._getTooltipShowClass();
 				if (this._state.inDrag === false && this.alwaysShowTooltip !== true) {
-					this._removeClass(this.tooltip, 'in');
-					this._removeClass(this.tooltip_min, 'in');
-					this._removeClass(this.tooltip_max, 'in');
+					this._removeClass(this.tooltip, showClass);
+					this._removeClass(this.tooltip_min, showClass);
+					this._removeClass(this.tooltip_max, showClass);
 				}
 				this._state.over = false;
 			},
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -199,6 +199,7 @@ Options can be passed either as a data (data-slider-foo) attribute, or as part o
 | focus | bool | false | Focus the appropriate slider handle after a value change. |
 | labelledby | string,array | null | ARIA labels for the slider handle's, Use array for multiple values in a range slider. |
 | rangeHighlights | array | [] | Defines a range array that you want to highlight, for example: [{'start':val1, 'end': val2, 'class': 'optionalAdditionalClassName'}]. |
+| bs_version | string | 'bs3' | Switch between Boostrap 3 ('bs3') and Bootstrap 4 ('bs4') |
```
seiyria: camelcase

seiyria:
I don't know if this is _entirely_ accurate (@rovolution can correct me), but we don't want bootstrap as a dependency in any way. you can add the relevant tooltip classes to our css files. this will solve both of your queries.

Thanks for taking the time to tackle this (this is one of our most requested features).

I've left a small comment on the implementation. I also think it would be best if you just add a "bs3" or "bs4" class to the slider and then let the rest of the classes sort themselves out on their own - that will make this change much simpler overall.

madflow:
Maybe could elaborate what is meant by "but we don't want bootstrap as a dependency in any way" ?

If this is true the current codebase should already be functioning without Bootstrap (which is not the case when it comes to the Tooltips at least I believe).

If  the `bootstrap-slider` should just become `generic-slider` - then we do not need bs3 and bs4 altogether ;)

seiyria:
All I mean is that we copy the styles into bootstrap slider instead of having a hard dependency on them.

src/js/bootstrap-slider.js
```diff
@@ -647,14 +647,15 @@ const windowIsDefined = (typeof window === "object");
 				}, this);
 
 				// Undo inline styles and classes on tooltips
+				var positionClasses = this._getTooltipPositionClassMap();
```
rovolution: could you make this a `const` instead?

src/js/bootstrap-slider.js
```diff
@@ -1124,23 +1126,28 @@ const windowIsDefined = (typeof window === "object");
 					delete this.eventToCallbackMap[eventName];
 				}
 			},
+			_getTooltipShowClass: function() {
+				return  this.options.bs_version === 'bs4' ? 'show' : 'in';	
+			},
 			_showTooltip: function() {
+				var showClass = this._getTooltipShowClass();
```
rovolution: could you make this a const instead?

src/js/bootstrap-slider.js
```diff
@@ -1124,23 +1126,28 @@ const windowIsDefined = (typeof window === "object");
...
-					this._addClass(this.tooltip_min, 'in');
-					this._addClass(this.tooltip_max, 'in');
+					this._addClass(this.tooltip_min, showClass);
+					this._addClass(this.tooltip_max, showClass);
 					this.tooltip.style.display = 'none';
 				}
 				this._state.over = true;
 			},
 			_hideTooltip: function() {
+				var showClass = this._getTooltipShowClass();
```
rovolution: could you make this a const instead?

src/js/bootstrap-slider.js
```diff
@@ -1371,27 +1378,27 @@ const windowIsDefined = (typeof window === "object");
 
 					var offset_min = this.tooltip_min.getBoundingClientRect();
 					var offset_max = this.tooltip_max.getBoundingClientRect();
-
+					var positionClasses = this._getTooltipPositionClassMap();
```
rovolution: could you make this a const instead?

src/js/bootstrap-slider.js
```diff
@@ -1851,7 +1858,17 @@ const windowIsDefined = (typeof window === "object");
 			_toPercentage: function(value) {
 				return this.options.scale.toPercentage.apply(this, [value]);
 			},
+			_getTooltipPositionClassMap() {
+				var prefix = this.options.bs_version === 'bs4' ? 'tooltip-' : '';
```
rovolution: could you make this a const instead?

src/js/bootstrap-slider.js
```diff
@@ -1851,7 +1858,17 @@ const windowIsDefined = (typeof window === "object");
...
+				return {
+					right: `${prefix}right`,
+					left: `${prefix}left`,
+					top: `${prefix}top`,
+					bottom: `${prefix}bottom`,
+				};
+			},
 			_setTooltipPosition: function(){
+				var positionClasses = this._getTooltipPositionClassMap();
```
rovolution: could you make this a const instead?

seiyria:
Yes, in that case do that. I couldn't remember but I'll defer to that.

madflow:
I rebased and implemented the changes. Let me know if there is anything missing on my end.

src/js/bootstrap-slider.js
```diff
@@ -1851,7 +1858,17 @@ const windowIsDefined = (typeof window === "object");
 			_toPercentage: function(value) {
 				return this.options.scale.toPercentage.apply(this, [value]);
 			},
+			_getTooltipPositionClassMap() {
+				const prefix = this.options.bsVersion === 'bs4' ? 'tooltip-' : '';
```
rovolution: could you wrap this conditional in parenthesis? makes the code a bit more readable

```js
const prefix = (this.options.bsVersion === 'bs4') ? 'tooltip-' : '';
```

src/js/bootstrap-slider.js
```diff
@@ -1124,23 +1126,28 @@ const windowIsDefined = (typeof window === "object");
 					delete this.eventToCallbackMap[eventName];
 				}
 			},
+			_getTooltipShowClass: function() {
+				return  this.options.bsVersion === 'bs4' ? 'show' : 'in';
```
rovolution: same feedback: https://github.com/seiyria/bootstrap-slider/pull/819/commits/555e7fcc789bc430fcecbf80235cf4d0a01b2bfb#r172343511

madflow:
I implemented the changes and rebased. However: I thing I will have to revisit how tooltips work in Bootstrap 4: https://getbootstrap.com/docs/4.0/components/tooltips/

I think the correct Bootstrap class would be "bs-tooltip-{placement}" (I misread or botched it, dunno) and the third party "popper.js" should be included when we will go down the "Bootstrap 4" way.

AlwaysVbNet:
Any news on the compatibility issue? Any idea when the compatible version will be released?

seiyria:
As no commits have been made, nothing has changed yet.

madflow:
I did a short revisit - the Bootstrap Tooltips prepend a "bs-tooltip-" to the placement . I amended the PR for this. 

The biggest issue I am facing is the positioning. I can not really point out what has changed and this exceeds my hopes for "quick win" with this PR.  Sorry for creating this PR prematuraly.  

At the moment I would argue to close this. It only provides the bare minimum: that the tooltips show up...

madflow:
I created a plunkr with my latest code for illustration purposes: 

https://plnkr.co/edit/kYjQC196w46gf7WFVNjR?p=preview

madflow:
This should be superseeded by https://github.com/seiyria/bootstrap-slider/pull/856

Issue #824, opened 2018-03-22
Modifying value before _mouseup is called will cause _mouseup passing the wrong value to callbacks


purpose233:
If user sets the value before the ```_mouseup``` function is called,  ```_mouseup``` function will pass the modified value to user's callback function instead of the real value when ```mouseup``` event happened.
This problem happens in specific situation, like using slider to control video player which frequently calls ```setValue``` function. 
Calculating the value by event like what ```_mousedown``` and ```_mousemove``` do might be fine.

seiyria:
We'd accept a PR to change this behavior!

PR #825, opened 2018-03-23
Fix the bug of _mouseup passing wrong value.

purpose233:
Make _mouseup get value from event and perform the same as _mousemove and _mousedown. #824

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1661,10 +1661,15 @@ const windowIsDefined = (typeof window === "object");
 					}
 				}
 			},
-			_mouseup: function() {
+			_mouseup: function(ev) {
 				if(!this._state.enabled) {
 					return false;
 				}
+
+				var percentage = this._getPercentage(ev);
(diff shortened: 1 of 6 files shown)


seiyria:
thanks! we'd also like it if you could make the code pass our build. @rovolution anything else?

jespirit:
I've been looking into this but I can't seem to find an issue with this.

I see where you can just recalculate the value on the "mouseup" event as you would with the "mousemove" and "mousedown", but is it necessary? The last value the slider has on the very last "mousemove" event **should** match the value of the slider on the "mouseup" event.

I have a JSFiddle where I check if there's a difference between the values when the slider is in motion ("mousemove" and "change" events), and when the slider stops which correspond to "mouseup" and "sliderStop" events, and there doesn't seem to be a difference.

In this case, it's difficult to write a test. Any suggestions?

https://jsfiddle.net/8rL7qo2f/

rovolution:
@jespirit it is failing the CI check though

PR #828, opened 2018-04-18
Fix determining mouse coords when pageXOffset >= 0

### LINKED ISSUE #594: The window has scroll,the slider calculate goes wrong
# Issues

I found that when i open f12 for debugger,the slider close to the right of the document,mouse event goes to a wrong position;
then i read node in bootstrap-slider.js,
in line 1392,the _getPercentage function calculate the slide move use mouse pos pageX,pageY,which makes the bug.
pageX:The mouse position relative to the left edge of the document. when i scroll from left to right ,the value changed.
The relative distance used to calculate the slide moves should based on an unchanged value,relative to the screen.

The solution:
**In line 573 and 580,I changed the mousePos from pageX,pageY to clientX,clientY,it works.**

x-yuri:
Fixes #594 

Should I tidy up the rest of the [spec file][sf]?

[sf]: https://github.com/seiyria/bootstrap-slider/pull/828/files#diff-73615cea1fe2d0803b81aebbac268f06

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -672,7 +672,7 @@ const windowIsDefined = (typeof window === "object");
 				} else {
 					this.stylePos = 'left';
 				}
-				this.mousePos = 'pageX';
+				this.mousePos = 'clientX';
 				this.sizePos = 'offsetWidth';
 			}
 			// specific rtl class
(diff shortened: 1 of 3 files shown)


seiyria:
Thanks! I think this is fine. @rovolution does it look good to you?

rovolution:
LGTM and thanks for the good unit test coverage @x-yuri! merging and releasing under 10.0.2

Issue #746, opened 2017-05-08
Selection of DOM around while sliding


etlel:
This is a great tool. I wonder if it is possible to prevent the selection of the DOM elements around the slider when sliding. The selection is a bit ticklish and easily catches the contents of the tooltip and beyond.

seiyria:
Can you provide an example of what you mean? I don't quite follow.

etlel:
Of course. On these screenshots im sliding and the text around is selected, if I don't stay straight on the handle.

![capture du 2017-05-07 22-32-40](https://cloud.githubusercontent.com/assets/17115179/25788190/31092b9a-3375-11e7-8cf5-93ff60d67b90.png)

![capture du 2017-05-07 22-29-32](https://cloud.githubusercontent.com/assets/17115179/25788150/f69cb7ce-3374-11e7-816f-d6855712ded4.png)

seiyria:
The only thing we could really do is text-selection: none (or whatever) on the tooltip. I don't think I've ever seen this myself, so I think that's what I'd recommend you do.

str:
This also happens to me

![selection_001](https://user-images.githubusercontent.com/123817/27735024-be246738-5d5a-11e7-8837-9c1aa50f294b.png)

Fl0ux:
I can confirm this issue with Firefox at least.

Solution : 
```css
.tooltip {
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

ktsaou:
I have the same problem.
@Fl0ux solution works for the tooltip, but selection is still enabled for the rest of the page:

![peek 2017-11-19 19-08](https://user-images.githubusercontent.com/2662304/32993196-295c42ce-cd5d-11e7-8f58-f2eb26a13369.gif)

ktsaou:
btw, the problem is also reproducible at the project's home page:

![peek 2017-11-19 19-14](https://user-images.githubusercontent.com/2662304/32993255-f7a69206-cd5d-11e7-8764-5c54b5a4ec6d.gif)

ktsaou:
I tried chrome, firefox, opera, edge, safari. It happens only on firefox on both windows and linux.

PR #829, opened 2018-04-22
Fix selection of surrounding DOM elements while sliding in FF and IE

wulfsoft:
I fixed the issue described in #746.


seiyria:
please provide a jsfiddle of your fix

rovolution:
just pulled it down and ran locally and it works! Thanks for your contribution @wulfsoft!

PR #838, opened 2018-06-05
Change readme on alternative namespace

putrefactio:
Added more info on how the alternative namespace works. I misinterpreted it when using this lib.


seiyria:
Looked good to me, thanks

PR #844, opened 2018-07-01
Add note about Bootstrap 4 compatibility

anderso:
I suggest to add a note in the README about Bootstrap 4 compatibility until #689 is resolved because I think this is a blocker for most users.


rovolution:
merged. thanks for your contribution @anderso!

PR #849, opened 2018-07-18
Use pointer cursor on slider ticks

edemaine:
I'm now updating from a very old version of this package, and I found that this CSS rule got removed somehow.  In my application, all the possible options are marked with ticks, and the ticks are clickable.  Yet the `cursor: pointer` rule somehow only applies everywhere *except* the tick marks (probably because of a change in how the tick marks are `position`ed).  Thus I found it necessary to make the tick marks also have the `cursor: pointer` rule.

Here's a screenshot of my use-case.  Unfortunately, I can't get the screenshot to show the cursor, but it is correct.

image

### DIFF
--- a/src/less/rules.less
+++ b/src/less/rules.less
@@ -219,6 +219,7 @@
 }
 .slider-tick {
 	position: absolute;
+	cursor: pointer;
 	width: @slider-line-height;
 	height: @slider-line-height;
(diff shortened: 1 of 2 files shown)


rovolution:
merged and published to v10.2.0. thanks again for your contribution @edemaine!

edemaine:
Great, thanks!  Published version works well.

Issue #854, opened 2018-08-03
Range slider, min value over max value.


Alhubing:
Hello, i've noticed that if you create a range slider, step value set at 1, you can drag the min handle over the max handle for 1 step.

There you have the example.

http://jsfiddle.net/95chpj1d/13/

Yo have to drag the min handle, and when the min value gets to the max value, it can go 1 unit over the max.

You can check it in the console. At some point you can get an output like [0, 1] -> [1, 1] -> [2 ,1]

You can get the same behaviour if you drag the max handle over the min handle.

Thanks.

seiyria:
It was implemented this way intentionally, if I remember correctly it will switch the handles around when it makes sense to do so.

Alhubing:
The handles don't change until there is a difference of more than one unit.

[1,0] -> [0,0] => [1,0] => [0,2]

The correct behaviour would be

[1,0] -> [0,0] => [0,1] => [0,2]

If you give different colours to each handle you can see how the min handle goes over the max handle and they don't switch.

Thanks.

Alhubing:
Please, could you give a hint about this?
Thank you.

jespirit:
I was able to reproduce the bug. Check this JSFiddle https://jsfiddle.net/9L65wxa3/

* Range starts at [5, 7]
To reproduce:
1. Drag the the lower handle past the upper handle so the values are [7, 8]
2. Tab control to the upper handle and press Left arrow twice.
3. Result is [7, 8] => [7, 7] => [7, 6]

I have a fix which is a single line of code. I just need to add unit test for it.

PR #864, opened 2018-10-24
Fix Issue #854 - Range slider, min value over max value.

jespirit:
Pull Requests
=============

- [x] JSFiddle (or an equivalent such as CodePen, Plunker, etc) or screenshot/GIF with new feature or bug-fix

This pull request fixes https://github.com/seiyria/bootstrap-slider/issues/854. You can check the issue for my JSFiddle that is used to reproduce the bug.

I added 2 unit tests as well as making a minor edit with a unit test in the same file: `test/specs/DraggingHandlesSpec.js`.

I will say that I was getting this error `Fatal error: Callback must be a function` displayed at the very end in red text, after running the unit tests using `grunt test` command. I don't know why, thoughts?

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1659,6 +1659,9 @@ const windowIsDefined = (typeof window === "object");
 				this._setDataVal(val);
 				this._trigger('slideStop', val);
 
+				// No longer need 'dragged' after mouse up
+				this._state.dragged = null;
+
 				return false;
 			},
 			_calculateValue: function(snapToClosestTick) {
--- a/test/specs/DraggingHandlesSpec.js
+++ b/test/specs/DraggingHandlesSpec.js
@@ -1,41 +1,60 @@
 describe("Dragging handles tests", function() {
 	var testSlider;
+	var mouseEventArguments;
+	var tickOffsets;
+
+	beforeEach(function() {
+		// Create slider
+		testSlider = new Slider(document.getElementById("testSlider1"), {
+			ticks: [0, 1, 2, 3, 4, 5, 6],
+			value: [4, 5],
+			step: 1,
+			range: true,
+		});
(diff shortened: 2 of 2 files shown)


src/js/bootstrap-slider.js
```diff
@@ -1659,6 +1659,9 @@ const windowIsDefined = (typeof window === "object");
 				this._setDataVal(val);
 				this._trigger('slideStop', val);
 
+				// No longer need 'dragged' after mouse up
+				delete this._state.dragged;
```
rovolution: generally im not a big fan of removing properties with the `delete` keyword (it has dangling memory reference issues in multiple browsers).

Can you refactor this to just set dragged to `null` or something like that instead?

rovolution:
@seiyria personally, im not too picky about fixing the code climate issue, and have debated just removing it altogether since i feel it doesnt deliver any real value TBH.

seiyria:
@rovolution That's fine with me. I'd rather replace it with a linter if we had that set up or some-such. Or just not have it altogether.

rovolution:
we already have a linter setup in our CI check - https://github.com/seiyria/bootstrap-slider/blob/master/Gruntfile.js#L274-L278

src/js/bootstrap-slider.js
jespirit: I was just copying the behaviour of the keydown event, that will set and delete `this._state.keyCtrl`. I will set it to null, it doesn't seem to break any of the logic.

https://github.com/seiyria/bootstrap-slider/blob/2272d5a0249b6ccb05efc2d5eced52720d8272aa/src/js/bootstrap-slider.js#L1565

rovolution:
@jespirit LGTM 👍 ill merge, publish, and update changelog.

rovolution:
@jespirit merged and published to v10.2.3.

PR #877, opened 2018-11-30
Fixes Issue #876: Values for range slider are not the same

### LINKED ISSUE #876: Values for range slider are not the same for 'slideStart' and 'change' events
When using the keyboard to move handles for a range slider at the min or max ends, the values for the `slideStart` and `change` events are different.

1. Open console.
2. Focus the keyboard on either handle and key to the left or right.

'slideStart' gives [-101, 0] while 'change' gives [-100, 0]

https://jsfiddle.net/jexrpwym/

Issues
=============
**NOTE:** Before filing a new issue, please search through the open issues list and check to see if an existing issue already exists that describes your problem. If so, leave a comment within that issue with the checklist items described below.

Please make sure you include the following in your issue report where appropriate:

- [x] JSFiddle (or an equivalent such as CodePen, Plunker, etc) with the an example demonstrating the bug
- [x] Instructions to replicate the bug in the Fiddle

jespirit:
Fixes #876 

I've added unit tests

### DIFF
--- a/test/specs/KeyboardSupportSpec.js
+++ b/test/specs/KeyboardSupportSpec.js
@@ -490,3 +490,67 @@ describe("Keyboard Support Tests", function() {
     });
   });
 });
+
+describe("Navigating range sliders with the keyboard", function() {
+  var mySlider;
+  var keyboardEvent;
+  var options;
+  var $handle1;
+  var $handle2;
+
+  describe("Range slider values", function() {
+
+    beforeEach(function() {
+      options = {
+        id: 'mySlider',
+        min: -100,
+        max: 100,
+        step: 1,
+        value: [0, 1],
+        range: true
+      };
+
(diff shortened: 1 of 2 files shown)


test/specs/KeyboardSupportSpec.js
```diff
@@ -489,4 +489,68 @@ describe("Keyboard Support Tests", function() {
...
+    afterEach(function() {
+      if (mySlider) {
+        if (mySlider instanceof Slider) { mySlider.destroy(); }
+        mySlider = null;
+      }
+    });
+
+    it("Should not go below minimum at 'slideStart'", function(done) {
+      options.value = [-100, 0];
+      mySlider = new Slider($('#testSlider1')[0], options);
```
rovolution: nice tests!

PR #881, opened 2018-12-01
Fix calculation of percentage values

jespirit:
The percentage values were being miscalculated.

Old equation was taking into account negative min values.

Wrong
`(val / this.options.max) * 100;`

Correct
`((val - this.options.min) / (this.options.max - this.options.min)) * 100;`

```
min, max = -100, 100
val = 10
// old
percentage
= (10 / 100) * 100
= 10%

// fix
percentage
= [ (10 - (-100)) / (100 - (-100) ] * 100
= [ (10 + 100) / (100 + 100) ] * 100
= 110 / 200 * 100
= 55%
```

So I just make call to `this._toPercentage()`.

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -1569,7 +1569,7 @@ const windowIsDefined = (typeof window === "object");
 				}
 
 				var val = this._state.value[handleIdx] + dir * this.options.step;
-				const percentage = (val / this.options.max) * 100;
+				const percentage = this._toPercentage(val);
 				this._state.keyCtrl = handleIdx;
 				if (this.options.range) {
 					this._adjustPercentageForRangeSliders(percentage);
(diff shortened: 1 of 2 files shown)


jespirit:
That is a difficult question. I'll see what I can do.

jespirit:
The percentage is used for checking if one handle overtakes another using either the keyboard or mouse and adjusts the percentages for both handles by switching them.

The code still seems to work even though the equation for calculating the percentage is incorrect. The percentage is incorrectly calculated based on a scale from 0 to `max`, regardless of the value for `min`.

```
// Current implementation
min,max = -100, +100
val, percentage
10, 10%
20, 20%
...
90, 90%
(percentages can be negative, which is obviously wrong, should be from 0 to 100)
-70, -70%
-60, -60%
...
```

**Note:** The percentages can be negative when the  `value` is negative.

However, the logic still works when checking if one handle overtakes another because the percentages are based on the scale from 0 to `max`, (eg. 70% < 90%) and even for negative numbers (eg. -70% < -60%).

As for the logarithmic scale, I have no idea. I can't say I know a lot about logarithmic scale.

I added two unit tests to check if the correct handles are receiving focus.

rovolution:
@jespirit thank you for doing that!!

PR #885, opened 2018-12-02
Fixes Issue #799: valuetext doesn't update

### LINKED ISSUE #799: valuetext doesn't update when formatted value is number-like
When using a formatter that conditionally returns a custom string, or otherwise returns the value as a string, the `aria-valuetext` isn't updated when switching to a value that doesn't use the custom format.

For example, the following formatter:
```
const formatter = v => (p(v) ? custom(v) : v.toString())
```

The cause is that that condition for whether to update `aria-valuetext` uses `isNaN` which returns `true` for strings that are number like, e.g., `"2017"`:

```
if (isNaN(this.options.formatter(this._state.value[0])) ) {
  this.handle1.setAttribute('aria-valuetext', this.options.formatter(this._state.value[0]));
}
```

There are a couple potential fixes:
1. Unset the value of `aria-valuetext` so there's no outdated information when the formatter returns a number-like value
2. Change the condition, e.g., to `typeof <formatted> === "string"` so that the text is always updated for formatters that return strings

Option 1 probably the better fix, unless you require that formatter always require a `string` return value

jespirit:
Fixes #799 

Pull Requests
=============

### DIFF
--- a/test/specs/AriaValueTextFormatterSpec.js
+++ b/test/specs/AriaValueTextFormatterSpec.js
@@ -53,5 +53,59 @@ describe("Aria-valuetext Tests", function() {
       var ttmaxMessage = $("#accessibilitySliderC").prev(".slider").children(".max-slider-handle").attr("aria-valuetext");
       expect(ttminMessage).toBe(expectedMessageC[0]);
       expect(ttmaxMessage).toBe(expectedMessageC[1]);
+
+      $('#accessibilitySliderC').slider('destroy');
+    });
+
+    describe("Unset 'aria-valuetext' attribute when value can be represented as a number", function() {
+      var $testSliderC;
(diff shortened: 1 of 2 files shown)


test/specs/AriaValueTextFormatterSpec.js
```diff
@@ -53,5 +53,59 @@ describe("Aria-valuetext Tests", function() {
...
+      var $testSliderC;
+      var dayOfWeek;
+
+      var dayFormatter = function(value) {
+        if (value[1]) {
+          return [ dayOfWeek[value[0]-1], dayOfWeek[value[1]-1] ];
+        }
+        return dayOfWeek[value-1];
+      };
+
```
rovolution: really good tests!!!!

PR #889, opened 2018-12-08
Fixes Issue #690 ticks should not overwrite min/max settings

### LINKED ISSUE #690: ticks should not overwrite min/max settings
## Problem:
Currently if you specify `ticks` the `min/max` settings get overwritten by the minimum and maximum tick value.

This causes a couple of problems:
1. Unexpected results if you have only one tick (at which point `min==max==ticks[0]`).
2. Prevents usage scenarios where ticks should not be at the minima/maxima of the slider.

https://jsfiddle.net/nonplus/ws68hrmf/

## Expected behavior:
1. If min configured, use specified min, regardless of ticks
1. If max configured, use specified max, regardless of ticks
1. If no min configured and ticks configured, use smallest tick as min (existing behavior)
1. If no max configured and ticks configured, use largest tick as max  (existing behavior)
1. If no min and no ticks configured, default min to 0  (existing behavior)
1. If no max and no ticks configured, default max to 10  (existing behavior)

I can provide a PR.

jespirit:
Fixes #690 

Pull Requests
=============

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -372,6 +372,9 @@ const windowIsDefined = (typeof window === "object");
 			options = options ? options : {};
 			var optionTypes = Object.keys(this.defaultOptions);
 
+			const isMinSet = options.hasOwnProperty('min');
+			const isMaxSet = options.hasOwnProperty('max');
+
 			for(var i = 0; i < optionTypes.length; i++) {
 				var optName = optionTypes[i];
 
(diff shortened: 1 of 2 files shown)


src/js/bootstrap-slider.js
```diff
@@ -372,6 +372,9 @@ const windowIsDefined = (typeof window === "object");
 			options = options ? options : {};
 			var optionTypes = Object.keys(this.defaultOptions);
 
+			var isMinSet = 'min' in options;
```
rovolution: could you refactor this to use `hasOwnProperty` instead of the `in` keyword?  `hasOwnProperty` checks the object only versus traversing the prototype chain

rovolution: also, refactor these `var`s to be `const`s (since their values do not change)

PR #902, opened 2018-12-22
makes event listener callbacks execute asynchronously

rovolution:
**Background**

After diving through our source code and tests, I realized a serious issue...none of our event listeners were executing asynchronously! 

Though we were exposing a callback based API, the callbacks were being executed synchronously whenever the private method `_trigger()` was being executed.

I wrapped the callback invocations in `setTimeout()` with a timeout of 0 (which cues the callback to be invoked on the next event loop tick) and fixed a bunch of our tests to reflect this.

Pull Requests
=============


test/specs/EventsSpec.js
```diff
@@ -131,167 +127,152 @@ describe("Event Tests", function() {
...
-          flag += 1;
+          done();
         });
         testSlider.slider('refresh');
         testSlider.data('slider')._mouseup(mouse);
-
-        expect(flag).toEqual(1);
       });
 
       it("slider should not bind multiple touchstart events after calling 'refresh'", function(done) {
```
jespirit: This test makes sure that 'slideStart' is not called more than once, hence the `flag` variable.
Using a spy: (I thought spies were confusing, but they're very useful).
```js
it("slider should not bind multiple touchstart events after calling 'refresh'", function(done) {
    flag = 0;
    var obj = {
        addOne: function() {
        flag++;
        }
    };
    spyOn(obj, 'addOne').and.callThrough();

    touch.initEvent("touchstart", true, true);

    testSlider.on('slideStart', obj.addOne);

    testSlider.slider('refresh');
    $('#testSlider2').prev('div.slider').find('.slider-handle').get(0).dispatchEvent(touch);

    window.setTimeout(function() {
        expect(obj.addOne.calls.count()).toBe(1);
        done();
    }, 0);
});
```

test/specs/EventsSpec.js
```diff
@@ -22,96 +22,92 @@ describe("Event Tests", function() {
 
     describe("Mouse Events", function() {
 
-      it("'slideStart' event is triggered properly and can be binded to", function() {
+      it("'slideStart' event is triggered properly and can be binded to", function(done) {
```
jespirit: There are lots of tests that "pass", but have no expectations and these log errors show up.

pull-902-no-expectations

test/specs/EventsSpec.js
```diff
@@ -22,96 +22,92 @@ describe("Event Tests", function() {
...
         testSlider.on('slideStart', function() {
-          flag = true;
+          done();
         });
         testSlider.data('slider')._mousedown(mouse);
-        expect(flag).toBeTruthy();
       });
 
-      it("'slide' event is triggered properly and can be binded to", function() {
+      it("'slide' event is triggered properly and can be binded to", function(done) {
```
jespirit: No expectations.

test/specs/EventsSpec.js
```diff
@@ -22,96 +22,92 @@ describe("Event Tests", function() {
...
           expect(e.value.toString()).toEqual(this.value);
+          done();
         });
         testSlider.slider("setValue", 3, true, false);
-
-        expect(flag).toBeTruthy();
       });
 
-      it("'change' event value and input value properties are synchronous", function() {
+      it("'change' event value and input value properties are synchronous", function(done) {
```
jespirit: No expectations... you get the point.

test/specs/DraggingHandlesSpec.js
```diff
@@ -221,28 +221,23 @@ describe("Dragging handles tests", function() {
         expect(testSlider.getValue()).toEqual([2, 5]);
     });
 
-    it("Should trigger change on mouseup", function() {
-        var ups = 0;
+    it("Should trigger change on mouseup", function(done) {
```
jespirit: Great! Fixes the recent pull request.

jespirit:
For future reference:

`setTimeout()` with delay of 0. Under **Late timeouts**:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified

test/specs/EventsSpec.js
rovolution: yea the reason i did this was that for these tests, there is nothing to really assert. i will see if there is an option we can pass to jasmine to suppress these

test/specs/EventsSpec.js
rovolution: https://github.com/seiyria/bootstrap-slider/pull/902/files#r243743659

test/specs/EventsSpec.js
rovolution: https://github.com/seiyria/bootstrap-slider/pull/902/files#r243743659

test/specs/EventsSpec.js
rovolution: cool i will update

test/specs/EventsSpec.js
rovolution: actually...i can refactor these to use the spy pattern we are using below. that way we can assert if the spy was invoked or not

rovolution:
@jespirit Thanks for the review! i updated with feedback. take a glance and let me know if it all looks good.

test/specs/EventsSpec.js
jespirit: Nice!

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
-          flag = true;
           expect(e.value.newValue.toString()).toEqual(testSlider.val());
+          done();
         });
         testSlider.slider("setValue", 3, false, true);
-
-        expect(flag).toBeTruthy();
       });
 
-      it("'slideStop' event is triggered properly and can be binded to", function() {
```
jespirit: ## $.trigger() is not async
I have removed the `done` argument and this test should fail because the 'slideStop' callback should be executed asynchronously. However, this test will still pass because `_trigger()` will call `$.trigger()` synchronously, and `.trigger()` will call custom events synchronously, with the exception of built-in events, which are executed sync/async depending on browser.

https://github.com/seiyria/bootstrap-slider/blob/a5fc012f717e31d292b8a837b2ba9ec9dcd8aa80/src/js/bootstrap-slider.js#L1876-L1883

See this thread: https://stackoverflow.com/questions/14017013/is-the-jquery-trigger-function-guaranteed-to-be-synchronous
```js
it("'slideStop' event is triggered properly and can be binded to", function() {
  touch.initEvent("touchstop");
  testSlider.on('slideStop', function() {  // this will be called synchronously
    expect(true).toBeTruthy();
  });
  testSlider.data('slider')._mouseup(mouse);
  window.setTimeout(function() {
    expect(spy).toHaveBeenCalled();
    // done();
  }, 0);
});
```

## Tests Should Be Async
Upon further review,

I think the issue is not that the callbacks are being executed synchronously, I think the way we had it before was fine, it's just that the **tests** themselves were not being executed asynchronously.

So I think the tests should be reorganized to be as follows:
```js
// Register callbacks
testSlider.on(/* slideStart, slide, slideStop */, function() { /* code here */});

setTimeout(function() {
  // set up code here to be executed asynchronously (next tick of the event loop)
}, 0);
```

For example,
```js
it("Should trigger change on mouseup", function(done) {
  var changes = 0;

  // Register callbacks
  testSlider.on('slideStop', function(){
    expect(changes).toBe(2);
    expect(testSlider.getValue()).toEqual([2, 5]);
    done();
  });
  
  testSlider.on('change', function(){
    changes++;
  });

  // Execute this code async (on next tick of the event loop)
  setTimeout(function() {
    testSlider.mousedown(mouseEvent('mousedown', tickOffsets[1]));
    expect(testSlider.getValue()).toEqual([1, 5]);

    testSlider.mouseup(mouseEvent('mouseup', tickOffsets[2]));
  }, 0);
});
```

If you remove the call to `done()` then the test will timeout.
```
Drag handles over each other and use keyboard to move handles over each other
     √ should drag and keydown handles properly to the right then back to the left
     √ should drag and keydown handles properly to the left then back to the right
   √ Should snap to a tick within tick bounds when using the mouse navigation
   × Should trigger change on mouseup
     Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. (1)
```

src/js/bootstrap-slider.js
```diff
@@ -1862,9 +1865,11 @@ const windowIsDefined = (typeof window === "object");
 
 				var callbackFnArray = this.eventToCallbackMap[evt];
 				if(callbackFnArray && callbackFnArray.length) {
-					for(var i = 0; i < callbackFnArray.length; i++) {
-						var callbackFn = callbackFnArray[i];
-						callbackFn(val);
+					for(let i = 0; i < callbackFnArray.length; i++) {
+						let callbackFn = callbackFnArray[i];
+						window.setTimeout(() => {
+							callbackFn(val);
```
jespirit: See my review comment below, the callbacks should be executed synchronously at the moment they're triggered, which is how `$.trigger()` behaves.

rovolution: thanks for pointing this out!

I reflected on this a bit more and I agree with you. Also, when a user triggers a mouse or keyboard based sliding event, this automatically triggers the necessary callbacks in an async fashion via the browsers native event handling mechanism.

If a user directly changes the value via the `setValue()` call, then the event handlers should execute in a sync fashion (as you indicated).

I look at another popular slider library called [noUiSlider](https://github.com/leongersen/noUiSlider), and this is how they do it as well.

rovolution:
@jespirit just removed the changes + removed the default timeout amount of 0 in the unit tests (`setTimeout` defaults to 0 seconds when a specific timeout is not specified)

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
-          flag = true;
           expect(isNaN(testSlider.val())).not.toBeTruthy();
+          done();
         });
         testSlider.data('slider')._mousemove(mouse);
-        expect(flag).toBeTruthy();
       });
 
-      it("'slide' event value and input value properties are synchronous", function() {
+      it("'slide' event value and input value properties are synchronous", function(done) {
```
jespirit: This test is testing whether the `slide` event is triggered synchronously. The `done` callback should be removed.

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
         testSlider.slider("setValue", 3, true, false);
-
-        expect(flag).toBeTruthy();
       });
 
-      it("'change' event value and input value properties are synchronous", function() {
+      it("'change' event value and input value properties are synchronous", function(done) {
         testSlider.on('change', function(e) {
-          flag = true;
           expect(e.value.newValue.toString()).toEqual(testSlider.val());
```
jespirit: This test is testing whether the `change` event is triggered synchronously. The `done` callback should be removed.

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
           expect(isNaN(testSlider.val())).not.toBeTruthy();
+          done();
         });
         testSlider.data('slider')._mousemove(touch);
-
-        expect(flag).toBeTruthy();
       });
 
-      it("'slide' event value and input value properties are synchronous", function() {
+      it("'slide' event value and input value properties are synchronous", function(done) {
```
jespirit: This test is testing whether the `slide` event is triggered synchronously. The `done` callback should be removed.

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
           expect(e.value.toString()).toEqual(testSlider.val());
+          done();
         });
         testSlider.slider("setValue", 3, true, false);
-
-        expect(flag).toBeTruthy();
       });
 
-      it("'change' event value and input value properties are synchronous", function() {
+      it("'change' event value and input value properties are synchronous", function(done) {
```
jespirit: This test is testing whether the `change` event is triggered synchronously. The `done` callback should be removed.

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
+          done();
+        });
       });
 
-      it("slider should not have duplicate events after calling 'refresh'", function() {
-        flag = 0;
-        testSlider.on('slideStop', function() {
-          flag += 1;
-        });
+      it("slider should not have duplicate events after calling 'refresh'", function(done) {
```
jespirit: This test makes sure that `slideStop` handler is not called more than once, hence the `flag` variable.

```js
it("slider should not have duplicate events after calling 'refresh'", function(done) {
    flag = 0;
    var obj = {
        addOne: function() {
        flag++;
        }
    };
    spyOn(obj, 'addOne').and.callThrough();

    testSlider.on('slideStop', obj.addOne);

    testSlider.slider('refresh');
    testSlider.data('slider')._mouseup(mouse);

    window.setTimeout(function() {
        expect(flag).toBe(1);
        expect(obj.addOne.calls.count()).toBe(1);
        done();
    }, 0);
});
```

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
 
       afterEach(function() {
         testSlider.destroy();
       });
 
-      it("'slide' event is triggered only once per slide action", function() {
-        testSlider.on('slide', function() {
-          numTimesFired++;
-        });
+      it("'slide' event is triggered only once per slide action", function(done) {
```
jespirit: This test makes sure that 'slide' event handler is not called more than once, hence the `numTimesFired` variable.
```js
it("'slide' event is triggered only once per slide action", function(done) {
    var obj = {
        addOne: function() {
          numTimesFired++;
        }
    };
    spyOn(obj, 'addOne').and.callThrough();

    testSlider.on('slide', obj.addOne);
    testSlider._mousemove(mouse);

    window.setTimeout(function() {
        expect(numTimesFired).toBe(1);
        expect(obj.addOne.calls.count()).toBe(1);
        done();
    }, 0);
});
```

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
+          expect(spy).toHaveBeenCalled();
+          done();
+        }, 0);
       });
 
-      it("'slideStart' event is triggered only once per slide action", function() {
-        testSlider.on('slideStart', function() {
-          numTimesFired++;
-        });
+      it("'slideStart' event is triggered only once per slide action", function(done) {
```
jespirit: Ditto

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
+          expect(spy).toHaveBeenCalled();
+          done();
+        });
       });
 
-      it("'slideStop' event is triggered only once per slide action", function() {
-        testSlider.on('slideStop', function() {
-          numTimesFired++;
-        });
+      it("'slideStop' event is triggered only once per slide action", function(done) {
```
jespirit: Ditto

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
+          expect(spy).toHaveBeenCalled();
+          done();
+        });
       });
 
-      it("'change' event is triggered only once per value change action", function() {
-        testSlider.on('change', function() {
-          numTimesFired++;
-        });
+      it("'change' event is triggered only once per value change action", function(done) {
```
jespirit: Ditto

test/specs/EventsSpec.js
```diff
@@ -21,325 +21,355 @@ describe("Event Tests", function() {
...
+      beforeEach(function() {
+        spy = jasmine.createSpy('spy');
       });
 
-      it("'slide' event is triggered properly and can be binded to", function() {
-        testSlider.on('slide', function() {
-          flag = true;
+      it("'slideStart' event is triggered properly and can be binded to", function(done) {
+        testSlider.on('slideStart', spy);
+        testSlider.data('slider')._mousedown(mouse);
```
jespirit: The following code is not good practice. The unit tests should avoid making calls to private methods such as `_mousedown()` or `_mouseup` etc. The proper solution is to use a `dispatchEvent()` call instead. However, to save you the work, this can be fixed in another pull request because this pattern occurs throughout this test file and others.
```js
testSlider.data('slider')._mousedown(mouse);
```

rovolution:
@jespirit feel free to make the changes you want if you don't want me to block you. i probably will not have time to work on this

test/specs/EventsSpec.js
jespirit: I'll address this in another pull request (when I get time).

Issue #916, opened 2019-03-12
$(el).slider() returns undefined if target element does not exist in DOM


ollycross:
Fiddle: https://jsfiddle.net/aduw6L0p/

Bootstrap Slider v10.6.1
Jquery v3.3.1
Tested using Chrome v72

__Description__
If you call `slider()` on a jQuery object which has matched no elements, it will return `undefined` which will break chaining.

__Steps to reproduce__
- Call `slider()` on a jQuery object which contains no elements
- Examine the returned value

__Expected behaviour__
The return value should be an empty jQuery object, the same as if you had called `$()`.  This would allow safe chaining with further jQuery functions such as `on()` regardless of whether the element has been matched or not.

__Actual behaviour__
The return value is `undefined` and calling other jQuery functions will throw an error.  The original object is unchanged (i.e. what you called `.slider()` on is unchanged.

ollycross:
Should note - as a workaround (as demonstrated in the fiddle) you can still attach events without chaining them:

```
const $el = $('#iDontExist');
$el.slider() // returns undefined
$el.on('slide', handleSlide);
```
Works fine, whether or not `#iDontExist` is present in DOM.

ollycross:
Upon investigation this may be an error with `Bridget` implementation in `bootstrap-slider`.

I'm not sure how you handle this dependency as it is bundled into the main `src/bootstrap-slider.js` file.  I've created a pull request which fixes the issue: https://github.com/seiyria/bootstrap-slider/pull/917

jespirit:
Issue fixed. Closing.

PR #917, opened 2019-03-12
fix bridget bug with empty jquery object

ollycross:
Addresses #916

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -185,11 +185,10 @@ const windowIsDefined = (typeof window === "object");
 			        return $(this);
 			      });
 
-			      if(!objects || objects.length > 1) {
-			      	return objects;
-			      } else {
+			      if(objects.length === 1) {
 			      	return objects[0];
 			      }
+			      return objects;
 			    }
 			  };
 
--- a/test/specs/PublicMethodsSpec.js
+++ b/test/specs/PublicMethodsSpec.js
@@ -2,6 +2,21 @@ describe("Public Method Tests", function() {
   var testSlider;
 
   describe("slider constructor", function() {
+    describe("returns a jQuery object if it is called on a jQuery object with zero or more matching elements", function() {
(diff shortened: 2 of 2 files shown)


src/js/bootstrap-slider.js
```diff
@@ -185,11 +185,10 @@ const windowIsDefined = (typeof window === "object");
 			        return $(this);
 			      });
 
-			      if(!objects || objects.length > 1) {
-			      	return objects;
-			      } else {
+			      if(objects.length === 1) {
```
jespirit: Great work and the JSFiddle is always helpful.

Just a couple of things.

1. We only need changes to the source `bootstrap-slider.js` file in `src/js`.
1. You can discard the changes to the distribution files (everything under `dist/` folder) because those files are generated when a release is ready to be published to npm
1. Lastly, can you add a test in `test/specs/PublicMethodsSpec.js` file for this specific issue to prevent future releases/patches from reintroducing this bug?

**Note:** The check `!objects` is not necessary because `.map(callback)` will always return a jQuery object.

ollycross: Okay
- I've reverted the changes to the `dist` folder back to the `v10.6.1` tag
- I've added tests in  `test/specs/PublicMethodsSpec.js` to make sure that when called on a jQuery object with zero or more matches it will return a jQuery object.

jespirit:
Check over the Javascript linting errors in the Travis CI build report. There are several errors complaining about variables that are not _defined_

ollycross:
Sorry, undefined var errors are now gone.

jespirit:
@ollycross Looks great!

Before I merge the changes, I just need you to squash your commits into 1 or 2 commits so that the commit history is easy to read and follow. Thanks!

ollycross:
Cool, all done.

jespirit:
Nice! Thanks for your effort!

rovolution:
@jespirit remember to publish this change and update the changelog thanks!

rovolution:
@jespirit no worries i will take care of this for now.

PR #953, opened 2021-01-22
preventOverflow flag to offset tooltips to stay within viewport

### LINKED ISSUE #649: tooltip is offset, out of the parent div
thanks for making this awesome slider.

I implemented the tooltip_split : true

the minimum one, the left is offset the div, so does the max one. please see the screenshot.
screenshot-local jack-2016-11-03-12-02-56
is it possible to make the tooltip  auto positioned when the reach their edge wether min or max to keep the tooltip visible? 

thanks.

### LINKED ISSUE #952: Tooltips via Popper.js?
Bootstrap.js (since v4) uses [Popper.js](https://popper.js.org/) to display tooltips.  One of the nice things about Popper.js is that has flags [to automatically prevent tooltips from leaving the screen](https://popper.js.org/docs/v2/modifiers/prevent-overflow/) and [to flip orientation of tooltips to do the same](https://popper.js.org/docs/v2/modifiers/flip/).  I've grown to really like these, and now my slider tooltips stick out for not having this feature -- especially that my left and right extremes tend to extend beyond the screen width (previously reported in #649).

It would be nice to be able to use Popper.js for slider tooltips.  I see a few possibilities:
* Add Popper.js as a dependency; hopefully not an issue for people using Bootstrap, though there are Popper v1 vs. v2 issues. (I believe Bootstrap v4 uses Popper v1 while Bootstrap v5 uses Popper v2, and there are some compatibility issues.)
* Detect Popper.js and use it if so (similar to current jQuery dependency).
* Add callback/option(s) for custom tooltips that would allow someone to switch to Popper.js (or another system) themselves.

Do you have a sense of which of these options (if any) would be most preferable from your perspective?

edemaine:
This PR fixes #649 and in some sense fixes #952.

I initially tried the second or third approach outlined in #952, but realized that it would be difficult to actually use Popper.js.  `createPopper` requires a DOM element to anchor the tooltip to, and in some uses of bootstrap-slider, there isn't a tick associated with the position being tooltipped.  We could create an artificial tooltip position DOM element to give to `createPopper`, but I instead tried a fourth approach:

4. Implement the logic of [`preventOverflow`](https://popper.js.org/docs/v2/modifiers/prevent-overflow/) manually.

This achieves the goal of no dependencies, adding a fairly small amount of code for nice functionality, hidden within a new `preventOverflow` option. I didn't change the defaults because it would break a lot of tests and backward compatibility, but happy to change.

I did not implement [flip](https://popper.js.org/docs/v2/modifiers/prevent-overflow/) yet, mainly because I care about it less. I could work on it if desired.

I've gotten this working well in my project (via `npm link`), which uses left-to-right horizontal sliders and always-shown tooltips.  Here are some screenshots

context | screenshot
---|---
left extreme | image
right extreme (including scrollbar) | image
in the middle | image

As this is my first contribution, I could use some guidance on how to test more thoroughly/broadly, and these PR requests:

### DIFF
--- a/src/js/bootstrap-slider.js
+++ b/src/js/bootstrap-slider.js
@@ -791,6 +791,11 @@ const windowIsDefined = (typeof window === "object");
 			// Bind window handlers
 			this.resize = this._resize.bind(this);
 			window.addEventListener("resize", this.resize, false);
+			// Simulate Popper's behavior of listening to both resize and scroll:
+			// https://popper.js.org/docs/v2/modifiers/event-listeners/
+			if (this.options.preventOverflow) {
+				window.addEventListener("scroll", this.resize, false);
+			}
 
 
 			// Bind tooltip-related handlers
@@ -906,6 +911,7 @@ const windowIsDefined = (typeof window === "object");
 				scale: 'linear',
 				focus: false,
 				tooltip_position: null,
+				preventOverflow: false,
 				labelledby: null,
 				rangeHighlights: []
 			},
@@ -1243,6 +1249,7 @@ const windowIsDefined = (typeof window === "object");
 				this._setText(this.tooltipInner, formattedTooltipVal);
 
 				this.tooltip.style[this.stylePos] = `${positionPercentages[0]}%`;
+				this._preventOverflow(this.tooltip);
 
 				function getPositionPercentages(state, reversed){
 					if (reversed) {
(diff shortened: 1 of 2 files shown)


seiyria:
Thank you for this! It's the first contribution we've gotten in a while so please be patient. @rovolution - what do you think? You spent more time on this so I want to get your opinion as well if possible.

edemaine:
Thanks for all the quick responses! Also, if you'd rather run with this yourself, tweaking it how you'd like/rather it be, feel free to do so (e.g. pushing to this branch).

src/js/bootstrap-slider.js
```diff
@@ -791,6 +791,11 @@ const windowIsDefined = (typeof window === "object");
 			// Bind window handlers
 			this.resize = this._resize.bind(this);
 			window.addEventListener("resize", this.resize, false);
+			// Simulate Popper's behavior of listening to both resize and scroll:
+			// https://popper.js.org/docs/v2/modifiers/event-listeners/
+			if (this.options.preventOverflow) {
+				window.addEventListener("scroll", this.resize, false);
```
rovolution, line 797: could you make sure this event listener is being removed in the `_removeSliderEventHandlers` cleanup method?

rovolution:
is there anyway to possibly add an automated test for this? i know sometimes it can be tricky with these types of changes

FezVrasta:
Popper supports virtual reference elements, it would allow you to position a tooltip even if there's not a html element to attach to.

PR #967, opened 2021-10-26
Update README.md

paullaros:


myccmj:
Really appreciate for your work,but I wonder if there's a way to move both the min-slider and max-slider when I drag the bar between them?

PR #980, opened 2024-10-02
Fixed deprecation warnings for division and darken/lighten

JorgenSolli:
These deprecation warnings should be fixed:
> Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

> Deprecation Warning: darken() is deprecated

> Deprecation Warning: lighten() is deprecated.

My apologies for not ticking any of the below boxes. I understand if this PR will be ignored. But at least I tried 😄 
We're using this library for work, and I don't have time allocated to set up Jasmine, run unit-tests, and provide examples.

Pull Requests
=============

### DIFF
--- a/src/sass/_rules.scss
+++ b/src/sass/_rules.scss
@@ -6,9 +6,9 @@
     width: $slider-horizontal-width;
     height: $slider-line-height;
     .slider-track {
-      height: $slider-line-height/2;
+      height: calc($slider-line-height/2);
       width: 100%;
-      margin-top: -$slider-line-height/4;
+      margin-top: - calc($slider-line-height/4);
       top:  50%;
       left: 0;
     }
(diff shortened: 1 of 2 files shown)


utilmind (dismissed):
ok

utilmind:
@JorgenSolli unfortunately "build" fails. Is it possible to fix the following warning?
In general I agree, everything make sense, but the following warning should be fixed.

image

JorgenSolli:
> @JorgenSolli unfortunately "build" fails. Is it possible to fix the following warning? In general I agree, everything make sense, but the following warning should be fixed.

Strange... it's valid syntax according to the SASS documentation.
I tried replicating it in my project, and it compiled fine. The only difference is that I'm using "sass" with Vite, while you're using "grunt-sass". I don't know if that matters, though. It's a bit outside of my technical knowledge.

I'm OK with reverting back to darken() and lighten() for now. Thoughts?

utilmind:
> I'm OK with reverting back to darken() and lighten() for now. Thoughts?

@JorgenSolli I'll check this out later today. In general your improvement is good and make sense to me + I agree that used tools (and probably grunt itself) are outdated. I'll try to build it in different way. If this fails, we'll have to keep darken/lighten.

